PR #46, opened 2018-08-14
Save Publishers calls

SergeyZhukovsky:


src/ledger_impl.cc
```diff
@@ -43,39 +45,74 @@ void LedgerImpl::CreateWallet() {
 }
 
 void LedgerImpl::OnLoad(const ledger::VisitData& visit_data) {
-
+  if (visit_data.domain.empty()) {
```
bridiver: do we want this to be in LedgerImpl? Seems like it would make more sense in bat publishers?

src/ledger_impl.cc
```diff
@@ -43,39 +45,74 @@ void LedgerImpl::CreateWallet() {
...
+    // Skip the same domain name
+    return;
+  }
+  visit_data_iter iter = current_pages_.find(visit_data.tab_id);
+  if (iter != current_pages_.end() && iter->second.domain == visit_data.domain) {
+    return;
+  }
+  if (last_shown_tab_id_ == visit_data.tab_id) {
+    last_tab_active_time_ = ledger_client_->GetCurrentTime();
```
bridiver: I was actually thinking that we want to change all of these methods to pass the timestamp in. The reason is that page loads may be starting while ledger is still initializing and we'll want to do something like this https://github.com/brave/brave-core/commit/56dbdfe2940d8923561e1db0525b0f7bbce6b0cb#diff-aa3505dbf36b5d03d8ba0751e0c99904R116 in the client when we call these methods

bridiver: if we delay sending them and they don't already have the timestamp info then it will be wrong when this finally gets called

include/bat/ledger/ledger.h
```diff
@@ -15,16 +15,17 @@
...
-  const std::string tld;
-  const std::string domain;
-  const std::string path;
-  const uint32_t tab_id;
-  uint64_t duration;
+  std::string tld;
+  std::string domain;
+  std::string path;
+  uint32_t tab_id;
+  //uint64_t duration;
```
bridiver: yea, we should just delete this, huh? Doesn't really make sense anymore

src/ledger_impl.cc
SergeyZhukovsky: ya probably it would be better to just add one more argument with a timestamp from a client in all functions

src/ledger_impl.cc
SergeyZhukovsky: well we can do that, but bat publisher was supposed to be used like an actual storage of accepted publishers, instead of doing logic of figuring out what event comes

src/ledger_impl.cc
```diff
@@ -43,39 +45,74 @@ void LedgerImpl::CreateWallet() {
...
+  }
+  current_pages_[visit_data.tab_id] = visit_data;
+  //LOG(ERROR) << "!!!LedgerImpl::OnLoad tab_id == " << visit_data.tab_id << ", domain == " << visit_data.domain
+  //  << ", tld == " << visit_data.tld << ", path == " << visit_data.path;
 }
 
 void LedgerImpl::OnUnload(uint32_t tab_id) {
-
+  //LOG(ERROR) << "!!!LedgerImpl::OnUnload tab_id == " << tab_id;
+  visit_data_iter iter = current_pages_.find(tab_id);
```
bridiver: I think we'll lose the page view here in the case where the tab is closed

src/ledger_impl.cc
bridiver: ok, I'm fine with that

src/ledger_impl.cc
SergeyZhukovsky: ok, I'll check that case

src/ledger_impl.cc
```diff
@@ -43,39 +45,74 @@ void LedgerImpl::CreateWallet() {
 }
 
 void LedgerImpl::OnLoad(const ledger::VisitData& visit_data) {
-
+  if (visit_data.domain.empty()) {
+    // Skip the same domain name
+    return;
+  }
+  visit_data_iter iter = current_pages_.find(visit_data.tab_id);
+  if (iter != current_pages_.end() && iter->second.domain == visit_data.domain) {
```
bridiver: this seems like it should maybe be a DCHECK? This condition should never happen if the rest of the code is correct, right?

src/ledger_impl.cc
bridiver: also shouldn't this always update `last_tab_active_time_`?

SergeyZhukovsky: we can have an OnLoad event from a tab that is not Shown right now and finished loading in the background, we will save it on OnShow event

src/ledger_impl.cc
SergeyZhukovsky: yeah, lets trigger that assert here

src/ledger_impl.cc
bridiver: ok, makes sense

src/bat/ledger/ledger.cc
```diff
@@ -11,6 +11,9 @@
 
 namespace ledger {
 
+VisitData::VisitData():
+    tab_id(0) {}
```
bridiver: this should default to -1 (`TabStripModel::kNoTab`)

src/ledger_impl.cc
```diff
@@ -24,7 +24,9 @@ LedgerImpl::LedgerImpl(ledger::LedgerClient* client) :
     bat_client_(new BatClient(this)),
     bat_publishers_(new BatPublishers(this)),
     bat_get_media_(new BatGetMedia(this)),
-    initialized_(false) {
+    initialized_(false),
+    last_tab_active_time_(0),
+    last_shown_tab_id_(0) {
```
bridiver: same as above `-1`

PR #51, opened 2018-08-21
Auto contributions, tipping, direct and recurring donation

SergeyZhukovsky:


include/bat/ledger/ledger.h
```diff
@@ -15,19 +15,53 @@
...
   uint32_t tab_id;
+  std::string local_month;
+  std::string local_year;
 };
 
+LEDGER_EXPORT struct PaidData {
+  PaidData();
+  PaidData(const std::string& _domain,
+           const double& _value,
+           const int64_t& _date,
```
bridiver: what is `date`? If it's an epoch timestamp I think we should change the name to make it more clear. Same thing as above applies to month and year?

include/bat/ledger/ledger.h
```diff
@@ -43,6 +77,8 @@ class LEDGER_EXPORT Ledger {
   virtual void CreateWallet() = 0;
   virtual void Reconcile() = 0;
 
+  virtual void MakePayment(const PaidData& paid_data) = 0;
+  virtual void AddRecurrentPayment(const std::string& domain_name, const double& value) = 0;
```
bridiver: `Recurrent` is also correct, but I think `Recurring` is more common usage for payments

include/bat/ledger/ledger.h
```diff
@@ -15,19 +15,53 @@
...
+
   std::string tld;
   std::string domain;
   std::string path;
   uint32_t tab_id;
+  std::string local_month;
+  std::string local_year;
 };
 
+LEDGER_EXPORT struct PaidData {
```
bridiver: maybe `Payment` or `PublisherPayment`? Either way I think we should start adding comments explaining what these are

include/bat/ledger/ledger.h
```diff
@@ -15,19 +15,53 @@
...
+};
+
 LEDGER_EXPORT struct VisitData {
   VisitData();
   VisitData(const std::string& _tld,
             const std::string& _domain,
             const std::string& _path,
-            uint32_t tab_id);
+            uint32_t _tab_id,
+            const std::string& _local_month,
```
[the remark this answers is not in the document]
bridiver: maybe what you're looking for here is the timezone?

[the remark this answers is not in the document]
SergeyZhukovsky: I assumed we will use ticks, but with the current implementation it doesn't matter if any of the client will use somethings, process ticks or timestamps as it will receive the delta when the page is going to be hidden. Month and year exactly the current time zone numbers.

include/bat/ledger/ledger.h
SergeyZhukovsky: It is a date when the payment(tipping, direct donation) was made

include/bat/ledger/ledger.h
```diff
@@ -15,19 +15,53 @@
 
 namespace ledger {
 
+LEDGER_EXPORT enum PUBLISHER_CATEGORY {
```
bridiver: these seem like payment types?

src/bat_publishers.cc
```diff
@@ -279,6 +359,37 @@ bool BatPublishers::isEligableForContribution(const ledger::PublisherInfo& info)
...
+
+  state_->monthly_balances_[GetBalanceReportName(year, month)] = report_balance;
+}
+
+void BatPublishers::getBalanceReport(const std::string& year,
+    const std::string& month, ledger::BalanceReportInfo& report_info) {
+  std::map<std::string, braveledger_bat_helper::REPORT_BALANCE_ST>::const_iterator iter = 
+    state_->monthly_balances_.find(GetBalanceReportName(year, month));
+  if (iter == state_->monthly_balances_.end()) {
+    return;
```
bridiver: shouldn't this return true/false?

include/bat/ledger/ledger.h
SergeyZhukovsky: Well we have VisitData structure for auto contributions, the structure was named the same way for direct payments

include/bat/ledger/ledger.h
SergeyZhukovsky: it should be seen for the client as we have to get diff lists based on categories

include/bat/ledger/ledger.h
SergeyZhukovsky: ok, I will correct

src/bat_publishers.cc
```diff
@@ -279,6 +359,37 @@ bool BatPublishers::isEligableForContribution(const ledger::PublisherInfo& info)
...
+  report_balance.earning_from_ads_ = report_info.earning_from_ads_;
+  report_balance.auto_contribute_ = report_info.auto_contribute_;
+  report_balance.recurring_donation_ = report_info.recurring_donation_;
+  report_balance.one_time_donation_ = report_info.one_time_donation_;
+
+  state_->monthly_balances_[GetBalanceReportName(year, month)] = report_balance;
+}
+
+void BatPublishers::getBalanceReport(const std::string& year,
+    const std::string& month, ledger::BalanceReportInfo& report_info) {
```
bridiver: I'm not sure if there is an actual code style guideline for this or not, but it's definitely more typical in the chromium code to use pointer notation instead of ref notation for values that will be updated by the method

src/bat_publishers.cc
SergeyZhukovsky: We can.

include/bat/ledger/ledger.h
bridiver: yes, but how it that date represented as an int64_t? If it's an epoch timestamp I think we should change date -> timestamp to be clear. To me `date` sounds like the day of the month, especially when there is also a month and year field

src/bat_publishers.cc
SergeyZhukovsky: is the library supposed to follow chromium style guideline?

include/bat/ledger/ledger.h
SergeyZhukovsky: ok, will do

include/bat/ledger/ledger.h
bridiver: `PaymentData` would probably be better then. `Paid` just doesn't make grammatical sense here.

include/bat/ledger/ledger.h
bridiver: but couldn't the same publisher be in both auto contribute and tipping or direct donation?

include/bat/ledger/ledger.h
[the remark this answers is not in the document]
bridiver: what happens if the visit starts in one month and ends in another?

src/bat_publishers.cc
bridiver: I think we should probably use the same standards everywhere

include/bat/ledger/ledger.h
SergeyZhukovsky: ok, I will correct

include/bat/ledger/ledger.h
SergeyZhukovsky: It could, we have it saved under diff keys in that case and under diff categories each.

src/bat_publishers.cc
SergeyZhukovsky: Well, there are lots of places like that already, I can change that one, but will it help for the past calls? We can probably follow that pattern for all feature commits.

include/bat/ledger/ledger.h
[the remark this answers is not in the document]
SergeyZhukovsky: Ya I thought about that, we still calculates it when it started

src/bat_publishers.cc
bridiver: Personally I would favor returning a unique_ptr and check for null instead of true/false to avoid creating a struct we might never actually use

include/bat/ledger/ledger.h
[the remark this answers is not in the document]
SergeyZhukovsky: Or we need to add more complexity by passing month and year on hiding and an actual time probably to calculate how long does the publisher spend there and here and put the publisher twice in diff reports for diff months

src/bat_publishers.cc
SergeyZhukovsky: It could be empty only when it's a wrong logic calls on clients. unique_ptr creates the var dynamically on a heap, stack allocation is a way faster. I always try to create vars on stack.

bridiver: btw - I believe the compiler should do return by value optimization if you just return the struct so that should be just as efficient (or possibly more?) than passing by reference

bridiver: there are two main cases that I see for passing non-const refs as params
1. If the method "returns" multiple values
2. If the "return" value will not always be set
typically case 2 would return a bool to inform the caller about whether or not the param value was set

include/bat/ledger/ledger.h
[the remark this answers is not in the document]
NejcZdovc: @bridiver every month we will reset publishers so visit can't go from one month to another. When new month starts, we will archive old month and start a new one from 0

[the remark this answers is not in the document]
bridiver: what we do with publishers and VisitData is independent of the fact that an actual visit can in fact start on August 31st and end on Nov 1. We can decide to attribute that entire visit to August, but it still spans two months

include/bat/ledger/balance_report_info.h
```diff
@@ -0,0 +1,28 @@
...
+#include "bat/ledger/export.h"
+
+namespace ledger {
+
+LEDGER_EXPORT struct BalanceReportInfo {
+  BalanceReportInfo();
+
+  double opening_balance_ = .0;
+  double closing_balance_ = .0;
+  double grants_avail_ = .0;
```
NejcZdovc: I would say to just name this `grants`, because someone could be confused with available suffix. This are not all available grants in that month, but claimed ones

include/bat/ledger/ledger.h
```diff
@@ -43,6 +77,8 @@ class LEDGER_EXPORT Ledger {
   virtual void CreateWallet() = 0;
   virtual void Reconcile() = 0;
 
+  virtual void MakePayment(const PaidData& paid_data) = 0;
```
NejcZdovc: I think that make payment should only accept publisherKey and category and we would figure out date stuff in `MakePayment`. User can't do payments for past months so it will only be a current month

include/bat/ledger/ledger.h
NejcZdovc: Can we change `domain_name` to `domain` so that is the same as in other functions

include/bat/ledger/ledger.h
```diff
@@ -60,16 +96,22 @@ class LEDGER_EXPORT Ledger {
 
   virtual void SetPublisherInfo(std::unique_ptr<PublisherInfo> publisher_info,
                                 PublisherInfoCallback callback) = 0;
-  virtual void GetPublisherInfo(const PublisherInfo::id_type& publisher_id,
+  virtual void GetPublisherInfo(const std::string& publisher_key,
                                 PublisherInfoCallback callback) = 0;
+  virtual void GetRecurrentDonationPublisherInfo(PublisherInfoCallback callback) = 0;
   virtual void GetPublisherInfoList(uint32_t start, uint32_t limit,
                                     PublisherInfoFilter filter,
+                                    ledger::PUBLISHER_CATEGORY category,
```
NejcZdovc: Note that tipping and direct donations should be returned as one list. All tables that we have are displaying it as one entity.

image

include/bat/ledger/ledger.h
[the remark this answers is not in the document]
SergeyZhukovsky: I think that it's ok to leave as it is right now, as in any case we will pay the contribution.

src/bat_publishers.cc
SergeyZhukovsky: agree, made like #2.

include/bat/ledger/balance_report_info.h
SergeyZhukovsky: ++

include/bat/ledger/ledger.h
SergeyZhukovsky: we cannot do it inside the lib, get date is a platform dependent function, we should move it outside of the lib and let client call it. The structure also contains a value.

NejcZdovc: so this should be then handled in payment service for brave-core?

include/bat/ledger/ledger.h
SergeyZhukovsky: ++

include/bat/ledger/ledger.h
SergeyZhukovsky: yep, I'll provide you links with samples from Android.

NejcZdovc: @SergeyZhukovsky thank you

include/bat/ledger/ledger.h
SergeyZhukovsky: Ya, I missed that part, added an ability to pull several categories at once.

src/ledger_impl.h
```diff
@@ -50,16 +50,23 @@ class LedgerImpl : public ledger::Ledger,
 
   void SetPublisherInfo(std::unique_ptr<ledger::PublisherInfo> publisher_info,
                         ledger::PublisherInfoCallback callback) override;
-  void GetPublisherInfo(const ledger::PublisherInfo::id_type& publisher_id,
+  void GetPublisherInfo(const std::string& publisher_key,
                         ledger::PublisherInfoCallback callback) override;
+  void GetRecurringDonationPublisherInfo(ledger::PublisherInfoCallback callback) override;
```
bridiver: `GetPublisherInfoList` already has a filter so I don't think we need another method here

include/bat/ledger/ledger.h
```diff
@@ -20,14 +20,40 @@ LEDGER_EXPORT struct VisitData {
...
   std::string domain;
   std::string path;
   uint32_t tab_id;
+  std::string local_month;
+  std::string local_year;
 };
 
+LEDGER_EXPORT struct PaymentData {
+  PaymentData();
+  PaymentData(const std::string& _domain,
```
bridiver: shouldn't this be a publisher_id?

include/bat/ledger/ledger.h
```diff
@@ -60,16 +88,22 @@ class LEDGER_EXPORT Ledger {
 
   virtual void SetPublisherInfo(std::unique_ptr<PublisherInfo> publisher_info,
                                 PublisherInfoCallback callback) = 0;
-  virtual void GetPublisherInfo(const PublisherInfo::id_type& publisher_id,
+  virtual void GetPublisherInfo(const std::string& publisher_key,
                                 PublisherInfoCallback callback) = 0;
+  virtual void GetRecurringDonationPublisherInfo(PublisherInfoCallback callback) = 0;
```
bridiver: `GetPublisherInfoList` already has a filter so I don't think we need another method here

include/bat/ledger/ledger.h
```diff
@@ -60,16 +88,22 @@ class LEDGER_EXPORT Ledger {
...
   virtual void GetPublisherInfoList(uint32_t start, uint32_t limit,
                                     PublisherInfoFilter filter,
+                                    int category,
```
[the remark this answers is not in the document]
bridiver: I probably shouldn't have started it off as an enum, but this was meant to handle all filterable params

include/bat/ledger/ledger.h
```diff
@@ -60,16 +88,22 @@ class LEDGER_EXPORT Ledger {
...
+                                    const std::string& month,
+                                    const std::string& year,
                                     GetPublisherInfoListCallback callback) = 0;
 
   virtual void SetPublisherMinVisitTime(uint64_t duration_in_milliseconds) = 0;
   virtual void SetPublisherMinVisits(unsigned int visits) = 0;
   virtual void SetPublisherAllowNonVerified(bool allow) = 0;
   virtual void SetContributionAmount(double amount) = 0;
+  virtual void SetBalanceReport(const std::string& year,
```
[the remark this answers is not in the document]
SergeyZhukovsky: It's unknown yet, but probably yes, the client will retrieve the balance and will fill that and pass through.

include/bat/ledger/ledger.h
SergeyZhukovsky: publisher_id is formed from _tld, we pass the domain here only

include/bat/ledger/ledger.h
SergeyZhukovsky: ok, we can move the category inside the filter as well as month and year and work with that.

include/bat/ledger/ledger.h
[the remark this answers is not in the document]
bridiver: it seems really odd to me that the client would fill out the balance report when all the info comes from the ledger library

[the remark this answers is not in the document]
SergeyZhukovsky: ya probably we will need to remove that, it's just for UI implementation right now, the endpoints are not fully ready yet

include/bat/ledger/publisher_info.h
```diff
@@ -37,6 +58,9 @@ LEDGER_EXPORT struct PublisherInfo {
   uint32_t percent;
   double weight;
   bool excluded;
+  std::string key;
```
bridiver: what is `key`?

include/bat/ledger/ledger.h
```diff
@@ -60,16 +88,22 @@ class LEDGER_EXPORT Ledger {
 
   virtual void SetPublisherInfo(std::unique_ptr<PublisherInfo> publisher_info,
                                 PublisherInfoCallback callback) = 0;
-  virtual void GetPublisherInfo(const PublisherInfo::id_type& publisher_id,
+  virtual void GetPublisherInfo(const std::string& publisher_key,
```
bridiver: is this now the full leveldb key? I think that should be hidden from the Ledger interface

include/bat/ledger/ledger.h
bridiver: what about YT and other pubs that all share a domain?

include/bat/ledger/ledger_client.h
```diff
@@ -57,6 +58,7 @@ class LEDGER_EXPORT LedgerClient {
                                 PublisherInfoCallback callback) = 0;
   virtual void LoadPublisherInfoList(uint32_t start, uint32_t limit,
                                     PublisherInfoFilter filter,
```
bridiver: this is no longer relevant here, correct?

include/bat/ledger/ledger.h
bridiver: for instance https://github.com/brave-intl/bat-native-ledger/pull/51/files#diff-bc89113382920b61b34c2e5e29d24acaR65 seems to conflate domain and publisher_id

include/bat/ledger/publisher_info.h
bridiver: since PublisherInfo is now effectively per month/year I think we should add month/year fields here too so the platform can tell them apart without having to understand the key

include/bat/ledger/publisher_info.h
```diff
@@ -19,10 +20,30 @@ LEDGER_EXPORT enum PublisherInfoFilter {
...
+
+LEDGER_EXPORT struct ContributionInfo {
+  ContributionInfo() {}
+  ContributionInfo(const double &value_, const uint64_t& date_):
+    value(value_),
+    date(date_) {}
+
+  std::string publisher;  // Filled only for recurrent donations
+  double value;
+  uint64_t date;
```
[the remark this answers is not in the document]
bridiver: maybe it should have all 3 like PaymentData?

include/bat/ledger/ledger.h
SergeyZhukovsky: That's what we receive from a client, the client basically tells us I want to give that publisher that amount. Do clients should know what is publisher_id? They just know what domain should they give a payment YT and twitch have channels ids in their URLs, we will probably increase that structure or add more fields when we add the support in the upcoming ticket

include/bat/ledger/publisher_info.h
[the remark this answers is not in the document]
SergeyZhukovsky: The structure looks like that:
`'3_2018_june.brave.com': {
                "contributions": [{value: 10,
                date: <uint64_t>}, {...}, ...]`. So there are lots of contribution in the current month and year.

include/bat/ledger/ledger_client.h
SergeyZhukovsky: we use it to retrieve publishers list

include/bat/ledger/publisher_info.h
SergeyZhukovsky: ok, agree

include/bat/ledger/ledger.h
bridiver: the client does have publisher_id because they get it from PublisherInfo, but maybe need some feedback from @NejcZdovc here

include/bat/ledger/ledger_client.h
bridiver: I just mean this param, not the whole method

include/bat/ledger/ledger.h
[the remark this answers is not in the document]
SergeyZhukovsky: I moved category, month and year inside the filter and made it structure.

include/bat/ledger/ledger_client.h
SergeyZhukovsky: the category, month and year moved inside filter.

include/bat/ledger/publisher_info.h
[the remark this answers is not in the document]
bridiver: understood, but shouldn't that information still be part of the contribution structure for the same reason that we are adding it to PublisherInfo?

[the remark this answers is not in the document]
SergeyZhukovsky: the ContributionInfo vector is inside PublisherInfo structure, we are adding month and year inside the PublisherInfo structure. Why do we need to duplicate it in each ContributionInfo structure?
https://github.com/brave-intl/bat-native-ledger/blob/split_publishers_list/include/bat/ledger/publisher_info.h#L62

include/bat/ledger/ledger.h
```diff
@@ -43,6 +69,8 @@ class LEDGER_EXPORT Ledger {
   virtual void CreateWallet() = 0;
   virtual void Reconcile() = 0;
 
+  virtual void MakePayment(const PaymentData& payment_data) = 0;
```
bridiver: I think in a follow-up PR we need to add a callback for this to handle errors on the client

SergeyZhukovsky: It's almost the same as OnLoad Onhide methods, they will call SetPublisherInfo function inside and there is a callback that publisher saved. We don't have an actual endpoint currently, so it will only notify if it saved successfully or not.

PR #55, opened 2018-08-28
Converts ms to s

NejcZdovc:


src/ledger_impl.cc
```diff
@@ -265,7 +265,7 @@ void LedgerImpl::GetPublisherInfoList(uint32_t start, uint32_t limit,
     prefix, callback);
 }
 
-void LedgerImpl::SetPublisherMinVisitTime(uint64_t duration) { // In milliseconds
+void LedgerImpl::SetPublisherMinVisitTime(unsigned int duration) { // In seconds
```
bridiver: this should still be uint64_t to match Time::InSeconds

include/bat/ledger/ledger.h
```diff
@@ -95,7 +95,7 @@ class LEDGER_EXPORT Ledger {
                                     const ledger::PublisherInfoFilter& filter,
                                     GetPublisherInfoListCallback callback) = 0;
 
-  virtual void SetPublisherMinVisitTime(uint64_t duration_in_milliseconds) = 0;
+  virtual void SetPublisherMinVisitTime(uint64_tb duration_in_seconds) = 0;
```
bridiver: tb?

NejcZdovc: yeah, fixes just now

bridiver: I don't think seconds or milliseconds really matters to ledger. As long as the client uses one or the other consistently everything should be correct either way.

bridiver: seconds is fine for now, but we could change it to `ticks` and let the client decide

NejcZdovc: yeah point of this PR is to sync it all across, so that we only use one type

PR #59, opened 2018-08-29
Refreshing publishers list periodically

gdregalo:
"Refreshing verified publishers list periodically" feature. 
Requires implementation of timers functionality on the client side

### DIFF
--- a/src/ledger_impl.cc
+++ b/src/ledger_impl.cc
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #include <ctime>
+#include <random>
 
 #include "ledger_impl.h"
 
@@ -177,7 +178,7 @@ void LedgerImpl::SavePublisherState(const std::string& data,
 
 void LedgerImpl::SavePublishersList(const std::string& data,
   ledger::LedgerCallbackHandler* handler) {
-  ledger_client_->SavePublishersList(data, handler);
+  ledger_client_->SavePublishersList(data, this);
 }
 
 std::string LedgerImpl::GenerateGUID() const {
@@ -528,7 +529,12 @@ void LedgerImpl::OnTimer(uint32_t timer_id) {
 
 
 void LedgerImpl::LoadPublishersListCallback(bool result, const std::string& response) {
-  const uint64_t retry_load_pub_list_sec = 300; //retry loading publishers list in 300 seconds if failed
+  std::random_device seeder;
+  const auto seed = seeder.entropy() ? seeder() : time(nullptr);
+  std::mt19937 eng(static_cast<std::mt19937::result_type> (seed));
+  std::uniform_int_distribution <> dist(300, 3600); //retry loading publishers list in 300-3600 seconds if failed
+  const uint64_t retry_load_pub_list_sec = dist(eng);
+
   uint64_t timer_offset{0ull};
 
   if (result && !response.empty()) {
@@ -555,6 +561,11 @@ void LedgerImpl::RefreshPublishersList() {
 
   //start timer
   ledger_client_->SetTimer(start_timer_in, last_pub_load_timer_id_);
+}
+
+void LedgerImpl::OnPublishersListSaved(ledger::Result result) {
+  assert(ledger::Result::OK == result);
+  bat_publishers_->OnPublishersListSaved(result);
 }
 
 }  // namespace bat_ledger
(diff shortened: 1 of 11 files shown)


src/ledger_impl.cc
```diff
@@ -502,4 +512,49 @@ void LedgerImpl::SetBalanceReport(const std::string& year,
...
+    handler_.AddRequestHandler(std::move(url_loader),
+      std::bind(&LedgerImpl::LoadPublishersListCallback,this,_1,_2));
+
+  }
+}
+
+
+
+void LedgerImpl::LoadPublishersListCallback(bool result, const std::string& response) {
+  const uint64_t retry_load_pub_list_sec = 300; //retry loading publishers list in 300 seconds if failed
```
bridiver: I think we want a longer and more random delay here to avoid DDoSing ourselves when there is a problem with the endpoint

src/ledger_impl.cc
```diff
@@ -159,6 +162,7 @@ void LedgerImpl::OnPublisherStateLoaded(ledger::Result result,
   }
 
   OnWalletInitialized(result);
+  RefreshPublishersList();
```
bridiver: maybe we should put this in `OnWalletInitialized`? Either way we only want to call it if `result == ledger::Result::OK`

bridiver: Also for privacy reasons I think we should persist the 48 interval somehow so we don't load the publisher list every time the browser opens. We don't want the download to be a proxy metric for browser usage

gdregalo: This is how it is now. The publisher list is reloaded every 48 hours (defined in static_values.h).
static const uint64_t _publishers_list_load_interval = 48 * 60 * 60; //48 hours in seconds

The value (const uint64_t retry_load_pub_list_sec = 300) is only used as retry attempt if initial load failed.

bridiver: right, my point about the retry is that if there is a problem with the endpoint, particularly if the problem persists for any period of time, we are going to have every client retrying at relatively short intervals. As the response time gets slower from more load, the problem for the endpoint is exacerbated because that creates even more overlap between the connection attempts. We're effectively designing in a DDoS attack on ourselves. Ideally I think we'd have a backoff algorithm, but at the very least we can have more variability in the retry (randomized with longer maximum retry time).

bridiver: I missed `PUBLISHER_STATE_ST.pubs_load_timestamp_` so that seems fine

bridiver: maybe a minimum of 5 minutes and a maximum of 1 hour for retry?

src/static_values.h
```diff
@@ -11,6 +11,9 @@
 #define LEDGER_PRODUCTION_SERVER        "https://ledger.mercury.basicattentiontoken.org"
 #define BALANCE_STAGING_SERVER          "https://balance-staging.mercury.basicattentiontoken.org"
 #define BALANCE_PRODUCTION_SERVER       "https://balance.mercury.basicattentiontoken.org"
+#define PUBLISHERSTAGING_SERVER         "https://publishers-staging.basicattentiontoken.org"
```
SergeyZhukovsky: Do we have production endpoint for publishers? Maybe it's better to add the definition here as well now to make an easier switch on it?

src/ledger_impl.cc
gdregalo: Good point about potential DDoS againt ourselves. 
I will make it random between 5-60 minutes

gdregalo:
Saving publishers list is inalienable part of the whole refreshing publisher list process now. 
It doesn't make sense to download the list without saving it. 
Saving a list can fail because of corrupted filesystem or database access synchronization bug introduced by us.
Both situations are very unlikely. 
Downloading the list and keep it in memory untill the application recover from above issues is not good approach.
It is because the problems above can exist forever and it also heavily complicates the logic.
If the app failed to download/save a verified publishers list, then it will retry the whole download/save process again in random interval 300-3600 seconds.

bridiver:
"Both situations are very unlikely." - this is the scaling problem that I've talked about before and is essentially a version of the lottery fallacy. While it may be unlikely for any given user, with a large number of users it will happen frequently. We made this mistake far too often in browser-laptop where people assumed that unlikely events probably wouldn't happen.

bridiver:
also, we will be keeping the list in memory anyway because we're converting it to a bloom filter

src/static_values.h
gdregalo: I don't know the answer. We can do it later.

PR #60, opened 2018-08-30
convert year to int

bridiver:
standardize ordering of category/month/year params
remove leveldb-specific publisher key

pre-req for sqlite see https://github.com/brave/brave-core/pull/241/commits/927b445ee72fd5de1282ca1e55b35587b36b42f9

### DIFF
--- a/include/bat/ledger/publisher_info.h
+++ b/include/bat/ledger/publisher_info.h
@@ -13,14 +13,15 @@
 namespace ledger {
 
 LEDGER_EXPORT enum PUBLISHER_CATEGORY {
-  ALL_CATEGORIES = 1 << 0,
   AUTO_CONTRIBUTE = 1 << 1,
   TIPPING = 1 << 2,
   DIRECT_DONATION = 1 << 3,
-  RECURRING_DONATION = 1 << 4
+  RECURRING_DONATION = 1 << 4,
+  ALL_CATEGORIES = (1 << 5) - 1,
 };
 
 LEDGER_EXPORT enum PUBLISHER_MONTH {
+  ANY = -1,
   JANUARY = 1,
   FEBRUARY = 2,
   MARCH = 3,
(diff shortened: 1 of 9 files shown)


include/bat/ledger/publisher_info.h
```diff
@@ -13,14 +13,15 @@
 namespace ledger {
 
 LEDGER_EXPORT enum PUBLISHER_CATEGORY {
-  ALL_CATEGORIES = 1 << 0,
   AUTO_CONTRIBUTE = 1 << 1,
   TIPPING = 1 << 2,
   DIRECT_DONATION = 1 << 3,
-  RECURRING_DONATION = 1 << 4
+  RECURRING_DONATION = 1 << 4,
+  ALL_CATEGORIES = (1 << 5) - 1,
```
SergeyZhukovsky: I don't understand why it is moved with such value to the end, what if we will have another categories in the future, it would look ugly.

include/bat/ledger/publisher_info.h
```diff
@@ -13,14 +13,15 @@
...
 };
 
 LEDGER_EXPORT enum PUBLISHER_MONTH {
+  ANY = -1,
```
SergeyZhukovsky: I would call it INVALID, first thing that was in my mind when I saw ANY is that there is a query to get ANY month

bridiver: it's meant to be a wildcard for month

include/bat/ledger/publisher_info.h
bridiver: because this actually marks all the bits as `1`. So you can do something like this
```
PUBLISHER_CATEGORY category = PUBLISHER_CATEGORY::ALL_CATEGORIES;
...
if (category & DIRECT_DONATION)
```

PR #61, opened 2018-08-31
Unification of build url

NejcZdovc:


src/bat_helper.h
```diff
@@ -401,6 +407,8 @@ namespace braveledger_bat_helper {
   std::string getMediaKey(const std::string& mediaId, const std::string& type);
 
   uint64_t getMediaDuration(const std::map<std::string, std::string>& data, const std::string& mediaKey, const std::string& type);
+
+  std::string buildURL(const std::string& path, const std::string& prefix = "", const SERVER_TYPES& server = SERVER_TYPES::LEDGER);
```
bridiver: `const SERVER_TYPES& server` is fine, but just for future reference there's no reason to pass a typical enum value by reference because the enum value is stored as an int

Issue #62, opened 2018-08-31
Dynamic production and verbose variables


NejcZdovc:
Flags shouldn't be static, we should be able to set them via setters. In client we would do that via flags

NejcZdovc:
for verbose mode let's do macro for vlog so we can trigger it the same way as we do it in core https://github.com/brave/brave-core/pull/241/commits/0a582dfb9314c87496abf7c14cd8b528de145fc7#diff-a5d16be9c9b52a6df7b866b1d75d3f5fR496

for production flag we should be able to provide flag when doing `npm run start`. If flag is not provided default value should be production

PR #63, opened 2018-08-31
Adds verbose mode

NejcZdovc:
Currently verbose mode is always set to true. That should be changed in https://github.com/brave-intl/bat-native-ledger/issues/62, where default would be then changed false.


NejcZdovc:
In native ledger we only cover response. Request in handled in client https://github.com/brave/brave-core/pull/241/commits/314e0f86091e69e45a4cb16c9a8d04fa4354f44f

PR #64, opened 2018-08-31
Yt twitch integration fixes

SergeyZhukovsky:


src/bat_helper.cc
```diff
@@ -1749,22 +1750,22 @@ namespace braveledger_bat_helper {
...
-  //     if (succeded) {
-  //       decoded.push_back((uint8_t)'\0');
-  //       braveledger_bat_helper::getJSONTwitchProperties((char*)&decoded.front(), parts);
-  //     }
-  //     else{
-  //       LOG(ERROR) << "getTwitchParts failed in getFromBase64";
-  //     }
-  //   }
-  // }
+  void getTwitchParts(const std::string& query, std::vector<std::map<std::string, std::string>>& parts) {
```
bridiver: I added params into OnXHRLoad so we could parse the query string using the chromium libs. Can we use that here? Seems safer to use the chromium parsing and decode

bridiver: https://github.com/brave/brave-core/blob/rewards/components/brave_rewards/browser/rewards_service_impl.cc#L262
I forgot to delete the TODO after implementing

bridiver: or add it to OnPostData if we're parsing from there?

include/bat/ledger/ledger.h
```diff
@@ -83,7 +87,16 @@ class LEDGER_EXPORT Ledger {
       uint32_t tab_id,
       const std::string& url,
       const std::map<std::string, std::string>& parts,
-      const uint64_t& current_time) = 0;
+      const std::string& first_party_url, 
+      const std::string& referrer,
+      const VisitData& visit_data) = 0;
```
bridiver: what goes in this VisitData that is different from the OnLoad call?

src/bat_get_media.cc
```diff
@@ -10,156 +10,161 @@
...
+      updated_visit_data.tld += TWITCH_PROVIDER_NAME;
+      updated_visit_data.favIconURL = media_publisher_info->favIconURL_;
+      uint64_t realDuration = getTwitchDuration(media_publisher_info->twitchEventInfo_, twitchEventInfo);
+      //LOG(ERROR) << "!!!realDuration == " << realDuration;
+      ledger::TwitchEventInfo oldInfo = media_publisher_info->twitchEventInfo_;
+      media_publisher_info->twitchEventInfo_ = twitchEventInfo;
+      media_publisher_info->twitchEventInfo_.status_ = getTwitchStatus(oldInfo, twitchEventInfo);
+
+      ledger_->SetMediaPublisherInfo(realDuration, std::move(media_publisher_info), updated_visit_data,
+        std::bind(&onVisitSavedDummy, _1, _2));
```
bridiver: do we want to get notified if this doesn't save successfully?

bridiver: maybe a TODO for some error handling?

src/ledger_impl.cc
```diff
@@ -220,7 +253,27 @@ void LedgerImpl::SetPublisherInfo(std::unique_ptr<ledger::PublisherInfo> info,
...
+      std::bind(&LedgerImpl::OnSetMediaPublisherInfo, this, duration, visit_data, callback, _1, _2));
+}
+
+void LedgerImpl::OnSetMediaPublisherInfo(const uint64_t& duration,
+                          const ledger::VisitData& visit_data,
+                          ledger::MediaPublisherInfoCallback callback,
+                          ledger::Result result,
+                          std::unique_ptr<ledger::MediaPublisherInfo> info) {
+  SaveMediaVisit(visit_data, duration);
+  callback(result, std::move(info));
```
bridiver: SaveMediaVisit is async so shouldn't this run when it actually completes?

src/bat_helper.cc
SergeyZhukovsky: the post data is based64 value and it's not really query, it's a json object like that `[{"event":"experiment_branch","properties":{"platform":"mobile_web","language":"en-us","gitHash":"964ae46c471c406f935380d04b8d8013904d88de","location":"channel","time":1535684252.833,"experiment_id":"aa19fae8-0e00-4922-8f1f-6699cc654671","experiment_name":"mobile_web_singapore_first_byte","experiment_group":"control","app_version":"964ae46c471c406f935380d04b8d8013904d88de","browser":"Mozilla/5.0 (Linux; Android) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.91 Mobile Safari/537.36 Brave","device_id":"ZTjPk6OrfHoCSCzSOcJ5hWwZgwExewqn","domain":"m.twitch.tv","host":"https://m.twitch.tv","app_session_id":"sUlnF57mQVSgT7z2Ewid1tlnAkX8I9d5","url":"https://m.twitch.tv/sypherpk","viewport_height":604,"viewport_width":412,"device_orientation":"portrait"}}]` and there are can be several events in one upload data

include/bat/ledger/ledger.h
SergeyZhukovsky: yep we consider it as independent visit

src/bat_get_media.cc
SergeyZhukovsky: probably it's just has to be removed, I'm going to remove it

src/ledger_impl.cc
SergeyZhukovsky: the callback is a callback above that we want to remove `onVisitSavedDummy`

src/bat_get_media.cc
SergeyZhukovsky: better to keep it probably as it goes the same way right now as other publishers, the same function call

src/bat_helper.cc
bridiver: I see, so you're splitting out the JSON chunks and then parsing?

src/ledger_impl.cc
bridiver: right, but either way don't we want to wait until SaveMediaVisit actually completes to run the callback? SaveMediaVisit might fail and even if it succeeds, it hasn't succeeded yet

src/bat_helper.cc
SergeyZhukovsky: it's just a one big chunk that's actually an array of objects, in most cases there is just one element in that array.

src/ledger_impl.cc
bridiver: I guess it's a general problem with saveVisit because it doesn't currently have any way to report success/failure so probably a separate issue we should file

SergeyZhukovsky: we don't need to wait any result at all as we don't need to do anything there, we can just remove that callback call

bridiver: can you put a TODO here and I'll create an issue to report success/failure from saveVisit?

bridiver: https://github.com/brave-intl/bat-native-ledger/issues/65

bridiver: I captured in the ticket above

include/bat/ledger/ledger_client.h
```diff
@@ -61,6 +64,10 @@ class LEDGER_EXPORT LedgerClient {
                                 PublisherInfoCallback callback) = 0;
   virtual void LoadPublisherInfo(PublisherInfoFilter filter,
                                 PublisherInfoCallback callback) = 0;
+  virtual void LoadMediaPublisherInfo(const std::string& publisher_id,
```
bridiver: I thought we wanted to go the other direction we these and tell the client what to save? This seems to do the opposite of the changes we want to make for PublisherInfo. Shouldn't we just pass the serialized json here?

bridiver: just like we do with state and previously did with publisher info before converting to sqlite

PR #66, opened 2018-09-01
Adds support for verified/excluded

NejcZdovc:


src/bat_helper.h
```diff
@@ -340,6 +340,21 @@ namespace braveledger_bat_helper {
     PUBLISHER
   };
 
+  struct PUBLISHER_SERVER_LIST {
```
bridiver: this name is somewhat confusing, maybe PUBLISHER_STATUS?

src/bat_helper.h
```diff
@@ -368,6 +383,8 @@ namespace braveledger_bat_helper {
 
   bool getJSONResponse(const std::string& json, unsigned int& statusCode, std::string& error);
 
+  bool getJSONPublisherServerList(const std::string& json, std::map<std::string, PUBLISHER_SERVER_LIST>& list);
```
bridiver: per above, the name makes it sound like this is a map of string -> some kind of list of publishers

src/bat_publishers.cc
```diff
@@ -36,7 +36,8 @@ namespace braveledger_bat_publishers {
 
 BatPublishers::BatPublishers(bat_ledger::LedgerImpl* ledger):
   ledger_(ledger),
-  state_(new braveledger_bat_helper::PUBLISHER_STATE_ST) {
+  state_(new braveledger_bat_helper::PUBLISHER_STATE_ST),
+  publisher_lists_(new braveledger_bat_helper::PUBLISHER_LISTS_ST) {
```
bridiver: to avoid confusion with other publisher info lists, maybe verified_publisher_list?

src/bat_publishers.cc
```diff
@@ -358,13 +366,28 @@ std::vector<braveledger_bat_helper::PUBLISHER_ST> BatPublishers::topN() {
 }
 
 bool BatPublishers::isVerified(const ledger::PublisherInfo::id_type& publisher_id) {
-  // TODO - implement bloom filter
-  return true;
+  if (!publisher_lists_ || publisher_lists_->server_list_.empty()) {
```
bridiver: seems like the logic for isVerified and isExcluded should be encapsulated in PUBLISHER_LISTS_ST

src/bat_helper.h
```diff
@@ -340,6 +340,21 @@ namespace braveledger_bat_helper {
     PUBLISHER
   };
 
+  struct PUBLISHER_SERVER_LIST {
+    bool verified;
+    bool excluded;
+  };
+
+  struct PUBLISHER_LISTS_ST {
```
bridiver: same as below, maybe verified publisher list? I think it's easy to get things confused right now

src/bat_publishers.cc
NejcZdovc: it's not just verified. see comment bellow

src/bat_helper.h
NejcZdovc: see below

NejcZdovc:
so plan for `PUBLISHER_LISTS_ST` is that we will have different list. First one is server list, where we get data about publishers (verified, excluded, site banner, donations, etc). Second one will be deleted list (will implement today) where we will add publishers id that user removed via UI. And maybe some more. So `PUBLISHER_LISTS_ST` can't be renamed in verified list, because it contain a  lot of different data in lists.

src/bat_helper.h
NejcZdovc: it's not just verified and exclude. see comment bellow

NejcZdovc:
plan for `PUBLISHER_LISTS_ST`
```
{
  server_list_: {
     verified: bool
     excluded: bool
     donate: {
        title: string
        description: string
        backgroundUrl: string
        logoUrl: string
        donationAmounts: double[]
  },
  deleted_list_: {
     [publisherk_key: string]: bool
  }
}
```

src/bat_publishers.cc
NejcZdovc: so you would send `publisher_lists_` into this two function and move it into bat_helper?

bridiver:
why store a separate list for deleted publishers instead of just marking them as deleted? They are easy to filter out now that the publisher info is stored in the db. Added a bunch of separately stored lists seems to just complicate things.

bridiver:
it also seems strange to me to create a data structure to hold lists that we don't have yet when we know it will be incompatible with the planned format for the list we do have right now. Wouldn't it make more sense to build an abstraction layer for the verified publishers list that will allow us to switch to the bloom filter without changing anything else?

NejcZdovc:
@bridiver I am redoing this PR

NejcZdovc:
@bridiver it's ready for a re-review

include/bat/ledger/publisher_info.h
```diff
@@ -36,6 +36,12 @@ LEDGER_EXPORT enum PUBLISHER_MONTH {
   DECEMBER = 12
 };
 
+LEDGER_EXPORT enum PUBLISHER_EXCLUDE {
+  DEFAULT = 0,
```
bridiver: I'm not sure I understand what `DEFAULT` is. Is this based on some other setting? Why isn't the default `false`?

src/bat_publishers.cc
```diff
@@ -448,8 +473,15 @@ std::vector<ledger::ContributionInfo> BatPublishers::GetRecurringDonationList()
...
-void BatPublishers::RefreshPublishersList(const std::string & pubs_list) {
-  ledger_->SavePublishersList(pubs_list);
+void BatPublishers::RefreshPublishersList(const std::string& json) {
+  ledger_->SavePublishersList(json);
+
+  std::map<std::string, braveledger_bat_helper::SERVER_LIST> list;
+  bool success = braveledger_bat_helper::getJSONServerList(json, list);
+
+  if (success) {
+    server_list_ = std::make_unique<std::map<std::string, braveledger_bat_helper::SERVER_LIST>>(list);
```
bridiver: this is fine, but there's really no reason to make this a unique_ptr. It will actually be easier if it isn't because you won't need all the `!server_list_` checks, you can just check for `empty()`

include/bat/ledger/publisher_info.h
bridiver: @SergeyZhukovsky explained it to me, but can we add some comments here for reference?

NejcZdovc: comments added

src/bat_publishers.cc

PR #67, opened 2018-09-01
fixed next interval calculation when downloading publishers list

gdregalo:


bridiver:
@NejcZdovc said the timer was running even when no ledger state was loaded?

NejcZdovc:
Issue created https://github.com/brave-intl/bat-native-ledger/issues/69

PR #71, opened 2018-09-02
Removes pinned

NejcZdovc:


NejcZdovc:
blocked on #66

bridiver:
this seems independent of #66 ?

PR #74, opened 2018-09-04
Adds support for get all reports

NejcZdovc:


NejcZdovc:
implemented in core https://github.com/brave/brave-core/pull/241/commits/3162763832864c501d187e98817773238131e6de

PR #75, opened 2018-09-04
Media refactor

NejcZdovc:


include/bat/ledger/ledger_client.h
```diff
@@ -30,7 +30,7 @@ LEDGER_EXPORT enum URL_METHOD {
 using PublisherInfoCallback = std::function<void(Result,
     std::unique_ptr<PublisherInfo>)>;
 using MediaPublisherInfoCallback = std::function<void(Result,
-    std::unique_ptr<MediaPublisherInfo>)>;
+    std::unique_ptr<PublisherInfo>)>;
```
SergeyZhukovsky: We can probably use the PublisherInfoCallback as it is the same

src/bat_publishers.cc
```diff
@@ -59,7 +59,7 @@ double BatPublishers::concaveScore(const uint64_t& duration) {
   return (std::sqrt(b2_ + a4_ * duration) - b_) / (double)a2_;
 }
 
-std::string getProviderName(const ledger::PublisherInfo::id_type publisher_id) {
+std::string getProviderName(const std::string publisher_id) {
```
SergeyZhukovsky: const std::string&

src/bat_publishers.cc
```diff
@@ -69,7 +69,7 @@ std::string getProviderName(const ledger::PublisherInfo::id_type publisher_id) {
   return "";
 }
 
-bool ignoreMinTime(const ledger::PublisherInfo::id_type publisher_id) {
+bool ignoreMinTime(const std::string publisher_id) {
```
SergeyZhukovsky: const std::string&

src/bat_get_media.cc
```diff
@@ -119,15 +119,22 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
...
+      std::map<std::string, ledger::TwitchEventInfo>::const_iterator iter = twitchEvents.find(media_key);
+      if (iter != twitchEvents.end()) {
+        oldEvent = iter->second;
+      }
+
+      ledger::TwitchEventInfo newEvent(twitchEventInfo);
+      newEvent.status_ = getTwitchStatus(oldEvent, newEvent);
+
+      uint64_t realDuration = getTwitchDuration(oldEvent, newEvent);
+      twitchEvents.emplace(media_key, newEvent);
```
SergeyZhukovsky: emplace doesn't replace the existing element, it just inserts new if the key doesn't exist, you never update the event value with that code, it's better to do `twitchEvents[media_key] = newEvent;`

src/bat_get_media.cc
```diff
@@ -148,13 +155,19 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
...
```
SergeyZhukovsky: the same as above

include/bat/ledger/ledger_client.h
NejcZdovc: done

src/bat_publishers.cc

src/bat_publishers.cc

src/bat_get_media.cc
NejcZdovc: done

src/bat_get_media.cc
NejcZdovc: done

PR #77, opened 2018-09-04
make production/verbose extern

bridiver:


bridiver:
see brave-core https://github.com/brave/brave-core/pull/392

PR #82, opened 2018-09-05
don't allow multiple concurrent init calls

bridiver:
actually fail on errors

### DIFF
--- a/src/ledger_impl.cc
+++ b/src/ledger_impl.cc
@@ -28,6 +28,7 @@ LedgerImpl::LedgerImpl(ledger::LedgerClient* client) :
     bat_publishers_(new BatPublishers(this)),
     bat_get_media_(new BatGetMedia(this)),
     initialized_(false),
+    initializing_(false),
     last_tab_active_time_(0),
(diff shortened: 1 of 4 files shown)


src/ledger_impl.cc
```diff
@@ -128,7 +136,7 @@ void LedgerImpl::OnXHRLoad(
     uint32_t tab_id,
     const std::string& url,
     const std::map<std::string, std::string>& parts,
-    const std::string& first_party_url, 
+    const std::string& first_party_url,
```
bridiver: my editor automatically removes trailing spaces, so that's all these changes are

src/ledger_impl.h
```diff
@@ -46,15 +46,15 @@ class LedgerImpl : public ledger::Ledger,
   std::string GenerateGUID() const;
   void Initialize() override;
   void Reconcile() override;
-  void CreateWallet() override;
+  bool CreateWallet() override;
```
bridiver: this now returns false if initialization is in progress

PR #86, opened 2018-09-06
Use probi for reports

NejcZdovc:


NejcZdovc:
brave core implementation https://github.com/brave/brave-core/pull/409

src/bignum.cc
```diff
@@ -0,0 +1,81 @@
...
+
+  bn_mul_basic(result, a, b);
+
+  bn_free(a);
+  bn_free(b);
+
+  return bigNumToString(result);
+}
+
+}  // namespace braveledger_bat_bignum
```
jasonrsadler: Need CRLF for EOF here.

src/bignum.cc
```diff
@@ -0,0 +1,81 @@
...
+  bn_read_str(big_num, probi.c_str(), probi.length(), 10);
+}
+
+std::string bigNumToString(bn_t& number) {
+  int result_length = bn_size_str(number, 10);
+  char result_char[result_length];
+  bn_write_str(result_char, result_length, number, 10);
+  std::string result_string(result_char);
+
+  bn_free(bn_t);
```
gdregalo: Not sure what `bn_free(bn_t)` does here. 
If intention was  `bn_free(number)`, then I don't think it's good idea to deallocate input in conversion function.
Sometimes you just need to convert it leaving input as it is.

NejcZdovc: sounds good, let me pull it out

src/bignum.cc
```diff
@@ -0,0 +1,81 @@
...
+namespace braveledger_bat_bignum {
+void prepareBigNum(bn_t& big_num, const std::string& probi) {
+  bn_null(big_num);
+  bn_new(big_num);
+  bn_read_str(big_num, probi.c_str(), probi.length(), 10);
+}
+
+std::string bigNumToString(bn_t& number) {
+  int result_length = bn_size_str(number, 10);
+  char result_char[result_length];
```
gdregalo: Variable length arrays are not supported by many compilers. Since we need to keep bat-native-ledger code portable, then it's better to use predefined length buffer for bn conversion:
const int MAX_BN_BUFF = 256; //global
char bn_str_buff[MAX_BN_BUFF];

src/bignum.cc
NejcZdovc: added

src/bignum.cc
NejcZdovc: done

src/bignum.cc
NejcZdovc: done

src/bignum.cc
```diff
@@ -0,0 +1,89 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
diracdeltas: should this have some tests?

NejcZdovc: added to the list https://github.com/brave/brave-browser/issues/1346

PR #101, opened 2018-09-16
User changing default monthly value is now indicated in state

### LINKED ISSUE #100: Default monthly value
1.  we need to know if user changed default monthly budget (add bool flag)
2. every time that we fetch balance we also get default monthly budget. If user did not change his monthly budget, we need to set it to the value returned in the balance response 
3. follow up to https://github.com/brave-intl/bat-native-ledger/pull/97: value should be only added if user changed it

Response
```
{
   "altcurrency":"BAT",
   "probi":"0",
   "balance":"0.0000",
   "unconfirmed":"0.0000",
   "rates":{
      "BTC":0.0000315,
      "ETH":0.000793831530483428,
      "XRP":0.6861360718870346,
      "BCH":0.0003652369718724429,
      "LTC":0.0035704742818971277,
      "DASH":0.0010850588712951685,
      "BTG":0.010842151526420773,
      "USD":0.23455291335672,
      "EUR":0.20217945114946598
   },
   "parameters":{
      "adFree":{
         "currency":"BAT",
         "fee":{
            "BAT":20
         },
         "choices":{
            "BAT":[
               10,
               15,
               20,
               30,
               50,
               100
            ]
         },
         "range":{
            "BAT":[
               10,
               100
            ]
         },
         "days":30
      }
   }
}
```

Default monthly budget:
`"fee":{"BAT":20}`

jasonrsadler:
Fixes #100 

When the user changes the default monthly value, `user_changed_state` is set to true.

### DIFF
--- a/include/bat/ledger/ledger.h
+++ b/include/bat/ledger/ledger.h
@@ -132,6 +132,7 @@ class LEDGER_EXPORT Ledger {
   virtual void SetPublisherAllowNonVerified(bool allow) = 0;
   virtual void SetPublisherAllowVideos(bool allow) = 0;
   virtual void SetContributionAmount(double amount) = 0;
+  virtual void SetUserChangedContribution() = 0;
   virtual void SetAutoContribute(bool enabled) = 0;
   virtual void SetBalanceReport(PUBLISHER_MONTH month,
                               int year,
--- a/src/bat_client.cc
+++ b/src/bat_client.cc
@@ -170,10 +170,19 @@ void BatClient::registerPersonaCallback(bool result,
   saveState();
 }
 
+bool BatClient::didUserChangeContributionAmount() const {
+  return state_->user_changed_fee_;
+}
+
 void BatClient::setContributionAmount(const double& amount) {
   state_->fee_amount_ = amount;
   saveState();
 }
+
+void BatClient::setUserChangedContribution() {
+    state_->user_changed_fee_ = true;
+    saveState();
+  }
 
 double BatClient::getContributionAmount() const {
   return state_->fee_amount_;
--- a/src/bat_client.h
+++ b/src/bat_client.h
@@ -31,12 +31,14 @@ class BatClient {
   void requestCredentialsCallback(bool result, const std::string& response);
   void registerPersonaCallback(bool result, const std::string& response);
   void setContributionAmount(const double& amount);
+  void setUserChangedContribution();
   void setAutoContribute(const bool& enabled);
   const std::string& getBATAddress() const;
   const std::string& getBTCAddress() const;
   const std::string& getETHAddress() const;
   const std::string& getLTCAddress() const;
   uint64_t getReconcileStamp() const;
+  bool didUserChangeContributionAmount() const;
   double getContributionAmount() const;
   bool getAutoContribute() const;
   bool isReadyForReconcile();
--- a/src/bat_helper.cc
+++ b/src/bat_helper.cc
@@ -507,6 +507,7 @@ static bool ignore_ = false;
     reconcileStamp_(0),
     settings_(AD_FREE_SETTINGS),
     fee_amount_(0),
+    user_changed_fee_(false),
     days_(0),
     auto_contribute_(true),
     rewards_enabled_(true) {}
@@ -523,6 +524,7 @@ static bool ignore_ = false;
     fee_currency_ = other.fee_currency_;
     settings_ = other.settings_;
     fee_amount_ = other.fee_amount_;
+    user_changed_fee_ = other.user_changed_fee_;
     days_ = other.days_;
     transactions_ = other.transactions_;
     ballots_ = other.ballots_;
@@ -553,6 +555,7 @@ static bool ignore_ = false;
         d.HasMember("fee_currency") && d["fee_currency"].IsString() &&
         d.HasMember("settings") && d["settings"].IsString() &&
         d.HasMember("fee_amount") && d["fee_amount"].IsDouble() &&
+        d.HasMember("user_changed_fee") && d["user_changed_fee"].IsBool() &&
         d.HasMember("days") && d["days"].IsUint() &&
         d.HasMember("transactions") && d["transactions"].IsArray() &&
         d.HasMember("ballots") && d["ballots"].IsArray() &&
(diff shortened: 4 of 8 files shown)


src/bat_helper.cc
```diff
@@ -964,6 +970,7 @@ static bool ignore_ = false;
       }
 
       parameters_days_ = d["parameters"]["adFree"]["days"].GetUint();
+      fee_amount_ = d["parameters"]["adFree"]["fee"]["BAT"].GetUint();
```
NejcZdovc: I think this needs to be `GetDouble`

src/ledger_impl.cc
```diff
@@ -433,6 +433,10 @@ void LedgerImpl::OnWalletProperties(ledger::Result result,
     info->balance_ = properties.balance_;
     info->rates_ = properties.rates_;
     info->parameters_choices_ = properties.parameters_choices_;
+
+    if (!bat_client_->getUserChangedContributionAmount()) {
+      info->fee_amount_ = properties.fee_amount_;
+    }
     if (std::find(info->parameters_choices_.begin(),
```
NejcZdovc: I think that this if should only happens when user changes his balance

src/ledger_impl.cc
```diff
@@ -433,6 +433,10 @@ void LedgerImpl::OnWalletProperties(ledger::Result result,
     info->balance_ = properties.balance_;
     info->rates_ = properties.rates_;
     info->parameters_choices_ = properties.parameters_choices_;
+
+    if (!bat_client_->getUserChangedContributionAmount()) {
+      info->fee_amount_ = properties.fee_amount_;
```
NejcZdovc: I think that we need to call `setContributionAmount` here as well, because we are using `CLIENT_STATE_ST` when fetching this info from JS and not `WALLET_PROPERTIES_ST`

src/ledger_impl.cc
jasonrsadler: Then, if balance is pulled from server, wouldn't monthly contribution be overwritten with default server value?

NejcZdovc: yes, idea is that contribute list and default value is set every time we get balance is user did not changed his balance manually. In this case we need to make sure that we have users value inside contribute list as well

NejcZdovc: well contribute list and default value are both returned from the server and will override previous values, so yes we change default value

src/ledger_impl.cc
NejcZdovc: server can change default value

jasonrsadler:
Implementation in brave-core

src/bat_client.h
```diff
@@ -31,12 +31,17 @@ class BatClient {
...
+  void setContributionAmount(double amount, 
+    const bool& userChanged);
   void setAutoContribute(const bool& enabled);
   const std::string& getBATAddress() const;
   const std::string& getBTCAddress() const;
   const std::string& getETHAddress() const;
   const std::string& getLTCAddress() const;
   uint64_t getReconcileStamp() const;
+  bool getUserChangedContributionAmount() const ;
+  bool etContributionAmount(double amount,
```
NejcZdovc: `etContributionAmount` -> `setContributionAmount`

src/bat_client.cc
```diff
@@ -170,11 +170,22 @@ void BatClient::registerPersonaCallback(bool result,
...
+
 void BatClient::setContributionAmount(const double& amount) {
   state_->fee_amount_ = amount;
   saveState();
 }
 
+void BatClient::setContributionAmount(double amount, 
+  const bool& userChanged) {
+    state_->fee_amount_ = amount;
+    state_->user_changed_fee_ = userChanged;
```
NejcZdovc: I think that this should be only set when `userChanged` is set to true. Otherwise you could override it with false. Once is set to true we shouldn't change it

src/bat_client.cc
```diff
@@ -170,11 +170,22 @@ void BatClient::registerPersonaCallback(bool result,
   saveState();
 }
 
+bool BatClient::getUserChangedContributionAmount() const {
```
NejcZdovc: Maybe this one should be named something like `didUserChangedContributionAmount`. When I first read it I though that I will get contribution amount that user changed

src/ledger_impl.cc
```diff
@@ -433,15 +438,20 @@ void LedgerImpl::OnWalletProperties(ledger::Result result,
     info->balance_ = properties.balance_;
     info->rates_ = properties.rates_;
     info->parameters_choices_ = properties.parameters_choices_;
-    if (std::find(info->parameters_choices_.begin(), 
-      info->parameters_choices_.end(), 
-      bat_client_->getContributionAmount()) == 
+
+    if (!bat_client_->getUserChangedContributionAmount()) {
+      info->fee_amount_ = properties.fee_amount_;
+      bat_client_->setContributionAmount(properties.fee_amount_);
```
NejcZdovc: I think that we should call `SetContributionAmount` from `LedgerImpl` here. Also this should pass `false` in.

src/bat_client.cc
jasonrsadler: One overload is set from server and this one for when user changes it. If for some unknown reason, we need to modify logic to specify `false` we would already have it. Can change inner to say `state_->user_changed_fee_ = true` and ignore `userChanged`. wdyt?

NejcZdovc: so what would you do?

jasonrsadler: We'll just only allow the flag to be true for now.

src/bat_helper.cc
```diff
@@ -507,9 +507,10 @@ static bool ignore_ = false;
     reconcileStamp_(0),
     settings_(AD_FREE_SETTINGS),
     fee_amount_(0),
+    user_changed_fee_(false) ,
```
NejcZdovc: nit: extra space

jasonrsadler:
Done

include/bat/ledger/ledger.h
```diff
@@ -132,6 +132,7 @@ class LEDGER_EXPORT Ledger {
   virtual void SetPublisherAllowNonVerified(bool allow) = 0;
   virtual void SetPublisherAllowVideos(bool allow) = 0;
   virtual void SetContributionAmount(double amount) = 0;
+  virtual void SetUserChangedContribution() = 0;
```
bridiver: I think this creates a confusing and error-prone api. If the goal is to be able to call SetContributionAmount internally without updating state_->user_changed_fee_ then the simplest change would be to set both in SetContributionAmount for the public api and use a different private api method to set fee_amount_ without also setting user_changed_fee_

PR #102, opened 2018-09-18
Checks if wallet exists

NejcZdovc:
this was created, because you can create wallet from the panel and settings and both of them needs to check if was created in the other place


NejcZdovc:
brave core implementation https://github.com/brave/brave-core/pull/466

PR #104, opened 2018-09-19
WIP: Adds excluded to PublisherInfoFilter

ryanml:


src/bat_publishers.cc
```diff
@@ -237,7 +244,8 @@ void BatPublishers::restorePublishers() {
   auto filter = CreatePublisherFilter("",
       ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE,
       ledger::PUBLISHER_MONTH::ANY,
-      -1);
+      -1,
+      ledger::PUBLISHER_EXCLUDE::EXCLUDED);
```
bridiver: it's not entirely clear what the filter means in this context. It reads like the filter would only return publishers that are excluded, but I don't think that's correct

bridiver: also, is there a "don't care" option that gets everything regardless of the setting?

include/bat/ledger/publisher_info.h
```diff
@@ -51,6 +51,7 @@ LEDGER_EXPORT struct PublisherInfoFilter {
   int category;
   PUBLISHER_MONTH month;
   int year;
+  PUBLISHER_EXCLUDE excluded;
```
bridiver: re: comment below, maybe we need to rename the var, the enum or both to make it more clear?

bridiver: or maybe it would be easier to understand if this was an `int` like category and allowed bitwise operations?

src/bat_publishers.cc
```diff
@@ -248,12 +256,9 @@ void BatPublishers::onRestorePublishersInternal(const ledger::PublisherInfoList&
...
 
   for (size_t i = 0; i < publisherInfoList.size(); i++) {
-    if (isExcluded(publisherInfoList[i].id,
-                   publisherInfoList[i].excluded)) {
-      // Set to PUBLISHER_EXCLUDE::DEFAULT (0)
-      setExclude(publisherInfoList[i].id,
-                 ledger::PUBLISHER_EXCLUDE::DEFAULT);
-    }
+    // Set to PUBLISHER_EXCLUDE::DEFAULT (0)
+    setExclude(publisherInfoList[i].id,
```
bridiver: is this right? We want to set all of them to DEFAULT?

src/bat_publishers.cc
ryanml: @bridiver that is the correct interpretation. This is inside `restorePublishers` so we want to reset the flag for publishers that are currently marked as excluded.

There isn't, we can add something like `PUBLISHER_EXCLUDE::ANY` like month that all pubs will fall under.

src/bat_publishers.cc
ryanml: @bridiver the queryset that feeds in to this function only contains excluded publishers that need to be restored.

src/bat_publishers.cc
bridiver: ok, I guess that also answers my question below then

PR #106, opened 2018-09-19
Grant captcha

NejcZdovc:


NejcZdovc:
core implementation https://github.com/brave/brave-core/pull/479

PR #107, opened 2018-09-21
Rewards panel

NejcZdovc:


SergeyZhukovsky:
What is windowId? Do you have a corresponding brave-core commit?

NejcZdovc:
@SergeyZhukovsky https://github.com/brave/brave-core/pull/459

include/bat/ledger/publisher_info.h
```diff
@@ -52,8 +52,8 @@ LEDGER_EXPORT struct PublisherInfoFilter {
   PUBLISHER_MONTH month;
   int year;
   PUBLISHER_EXCLUDE excluded;
-
   std::vector<std::pair<std::string, bool>> order_by;
+  unsigned int min_duration;
```
ryanml: Any reason this isn't `uint64_t`?

NejcZdovc: so this is min duration for visit time, which means that max value is 60 (1minute).

PR #109, opened 2018-09-21
Fixes filter order by

NejcZdovc:


src/bat_publishers.cc
```diff
@@ -372,6 +372,7 @@ void BatPublishers::synopsisNormalizer(const ledger::PublisherInfo& info) {
       ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE,
       info.month,
       info.year);
+  filter.order_by.push_back(std::pair<std::string, bool>("ai.percent", false));
```
SergeyZhukovsky: looks good, but that's again an example that we need to move all sqlite structures inside the lib. Currently the client should deal itself of saving the PublisherInfo structure, but here we explicitly sort via `ai.percent`.

PR #114, opened 2018-09-26
Youtube and Twitch videos now provide more accurate times than before

jasonrsadler:
Addresses brave/brave-browser#971

See brave/brave-core#522 for test plan

### DIFF
--- a/src/bat_get_media.cc
+++ b/src/bat_get_media.cc
@@ -6,6 +6,7 @@
 #include "bat_get_media.h"
 
 #include <sstream>
+#include <cmath>
 
 #include "bat_get_media.h"
 #include "bat_helper.h"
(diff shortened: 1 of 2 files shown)


include/bat/ledger/publisher_info.h
```diff
@@ -51,9 +51,8 @@ LEDGER_EXPORT struct PublisherInfoFilter {
   int category;
   PUBLISHER_MONTH month;
   int year;
-  PUBLISHER_EXCLUDE excluded;
-
   std::vector<std::pair<std::string, bool>> order_by;
+  PUBLISHER_EXCLUDE excluded;
```
ryanml: Can we flip this back? per: https://github.com/brave-intl/bat-native-ledger/pull/119

src/bat_get_media.cc
```diff
@@ -118,12 +113,14 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
           _2,
           _3));
     } else if (TWITCH_MEDIA_TYPE == providerName) {
-      const std::string mediaUrl = getMediaURL(new_media_id, providerName);
+      const std::string twitchMediaID = 
+        braveledger_bat_helper::split(mediaId, '_')[0];
```
ryanml: Should we do a substring check for `_`? Just so we aren't accessing an index that doesn't exist. Let me know if `braveledger_bat_helper::split` accounts for this

jasonrsadler: Done.

include/bat/ledger/publisher_info.h
jasonrsadler: Done.

PR #117, opened 2018-09-26
Adding numExcludedSites to publisher state

### LINKED ISSUE #115: Publishers state should keep track of the number of excluded sites
We need this to be a simple integer attribute of the publisher state: `num_excluded_sites`. This will be used by the auto contribute modal.

ryanml:
Fixes #115 

This adds `num_excluded_sites_` to the publisher state. This value is kept track of internally inside the native-ledger, thus it doesn't need any setter methods in the ledger implementation.

**This should be tested with:** https://github.com/brave/brave-core/pull/475

### DIFF
--- a/src/bat_publishers.cc
+++ b/src/bat_publishers.cc
@@ -179,6 +179,13 @@ void onVisitSavedDummy(ledger::Result result,
   // onPublisherInfoUpdated will always be called by LedgerImpl so do nothing
 }
 
+void BatPublishers::setNumExcludedSitesInternal(ledger::PUBLISHER_EXCLUDE exclude) {
+  unsigned int previousNum = getNumExcludedSites();
+  setNumExcludedSites((exclude == ledger::PUBLISHER_EXCLUDE::EXCLUDED)
+                      ? ++previousNum
+                      : --previousNum);
+}
+
 void BatPublishers::makePaymentInternal(
       ledger::PaymentData payment_data,
       ledger::Result result,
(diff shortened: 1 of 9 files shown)


NejcZdovc:
@ryanml does this already works with that PR?

ryanml:
> @ryanml does this already works with that PR?

Yep, ready for pairing with: https://github.com/brave/brave-core/pull/475

src/bat_publishers.cc
```diff
@@ -237,7 +246,7 @@ void BatPublishers::onSetExcludeInternal(ledger::PUBLISHER_EXCLUDE exclude,
   publisher_info->month = ledger::PUBLISHER_MONTH::ANY;
 
   ledger_->SetPublisherInfo(std::move(publisher_info),
-      std::bind(&onVisitSavedDummy, _1, _2));
+      std::bind(&BatPublishers::onPublisherExcludeUpdated, this, exclude, _1, _2));
```
NejcZdovc: we should keep this call as is and in the previous line call `onPublisherExcludeUpdated`. `onPublisherExcludeUpdated` don't need to have publisher info, so I would name it just `setExcludedNum` or something like this

src/bat_publishers.cc
```diff
@@ -275,9 +282,11 @@ void BatPublishers::onSetExcludeInternal(ledger::PUBLISHER_EXCLUDE exclude,
   publisher_info->year = -1;
   publisher_info->excluded = exclude;
   publisher_info->month = ledger::PUBLISHER_MONTH::ANY;
+  setNumExcludedSitesInternal(exclude);
 
   ledger_->SetPublisherInfo(std::move(publisher_info),
       std::bind(&onVisitSavedDummy, _1, _2));
+  OnExcludedSitesChanged();
```
NejcZdovc: should this be part of `setNumExcludedSitesInternal`?

NejcZdovc: well I guess not

PR #121, opened 2018-09-28
Legacy Niceware based passphrases can now be used to restore wallets

jasonrsadler:
Fixes brave/brave-browser#1193


src/bat_helper.cc
```diff
@@ -1932,4 +1933,44 @@ static bool ignore_ = false;
...
 
+  uint8_t niceware_mnemonic_to_bytes(const std::string& w,
+    std::vector<uint8_t>& bytes_out, size_t *written) {
+    std::vector<std::string> wordList = split(w, WALLET_PASSPHRASE_DELIM);
+    if (wordList.size() < 16) { // not a valid legacy wallet word list
+      return INVALID_LEGACY_WALLET;
+    }
+
+    std::string file = readFile
+      ("./src/brave/vendor/bat-native-ledger/niceware/wordlist");
```
jasonrsadler: @bridiver @SergeyZhukovsky Where should we be reading in plain-text files. This works under dev but expect it to fail on Release.

include/bat/ledger/ledger_callback_handler.h
```diff
@@ -36,6 +37,8 @@ class LEDGER_EXPORT LedgerCallbackHandler {
 
   virtual void OnLedgerStateLoaded(Result result,
                                    const std::string& data) {};
+  virtual void OnNicewareListLoaded(Result result,
```
bridiver: I know you're following some established patterns here, but these were created before we had the callback binding issues sorted out between the library and brave-core. Please pass a callback into the method (SavePublisherInfo is an example)

src/bat_client.cc
```diff
@@ -928,11 +928,42 @@ void BatClient::saveState() {
...
 
+std::vector<std::string> BatClient::GetWordDictionary() const {
+  return word_dictionary_;
+}
+
+void BatClient::SetWordDictionary(const std::vector<std::string>& wordlist) {
+  word_dictionary_ = wordlist;
+}
+
+bool BatClient::loadNicewareList(const std::string& data) {
```
bridiver: is there any reason not to just call `SetWordDictionary` directly and skip this extra method? Returning `true` from a simple setter doesn't really seem to add any value?

src/bat_client.cc
```diff
@@ -928,11 +928,42 @@ void BatClient::saveState() {
...
-  int result = bip39_mnemonic_to_bytes(nullptr, passPhrase.c_str(), &newSeed.front(), newSeed.size(), &written);
+  std::vector<unsigned char> newSeed;
+  // TODO(jasonrsadler): Mobile doesn't have rewards so the size() == 16 should
+  // just result in invalid.
+  if (braveledger_bat_helper::split(passPhrase,
+    WALLET_PASSPHRASE_DELIM).size() == 16)
+  {
+
+    // use niceware for legacy wallet passphrases
+    ledger_->SetPendingPhrase(passPhrase);
```
bridiver: with the callback method you can prebind passPhrase (side note- use underscore naming ->`pass_phrase`) to the callback and `SetPendingPhrase` will no longer be needed

src/bat_helper.cc
```diff
@@ -1932,4 +1932,29 @@ static bool ignore_ = false;
     return ignore_;
   }
 
+  uint8_t niceware_mnemonic_to_bytes(const std::string& w,
+    std::vector<uint8_t>& bytes_out, size_t *written, 
+    std::vector<std::string> wordDictionary) {
+    std::vector<std::string> wordList = split(w, WALLET_PASSPHRASE_DELIM);
+    if (wordList.size() < 16) { // not a valid legacy wallet word list
```
diracdeltas: should be `!= 16`

jasonrsadler: Fixed

src/test/niceware_partial_unittest.cc
```diff
@@ -0,0 +1,105 @@
...
+    ui::ResourceBundle::GetSharedInstance().GetRawDataResource(
+    IDR_BRAVE_NICEWARE_LIST).as_string(), DICTIONARY_DELIMITER);
+
+  int result = braveledger_bat_helper::niceware_mnemonic_to_bytes
+    (passPhrase, nSeed, &written, data);
+  
+  ASSERT_TRUE(result == 0 && 
+    written == NICEWARE_BYTES_WRITTEN &&
+    nSeed.size() == NICEWARE_BYTES_WRITTEN);
+}
```
diracdeltas: For the test cases where bytes are written, can you check the result against the actual bytes instead of just the size? To see the bytes you can load the word phrase into browser laptop's restore wallet feature, then add some logging to see what the wallet bytes are.

You could also get the bytes using my niceware tool: https://diracdeltas.github.io/niceware/

Also please add the test cases from https://github.com/diracdeltas/niceware/blob/master/test/mainTest.js#L56

include/bat/ledger/ledger_callback_handler.h
jasonrsadler: Fixed

src/bat_client.cc
jasonrsadler: No value. Fixed

src/bat_client.cc
jasonrsadler: Fixed

include/bat/ledger/ledger.h
```diff
@@ -169,7 +169,7 @@ class LEDGER_EXPORT Ledger {
                                const std::string& baseDomain,
                                const std::string& path,
                                ledger::PUBLISHER_MONTH month,
-                               int year) = 0;
+                               int year) = 0;
```
bridiver: this seems accidental? Don't really need to add extra whitespace here

src/bat_client.cc
```diff
@@ -928,11 +928,46 @@ void BatClient::saveState() {
...
+    continueRecover(result, &written, newSeed);
+  }
+}
+
+void BatClient::OnNicewareListLoaded(const std::string& pass_phrase,
+                                        ledger::Result result,
+                                        const std::string& data) {
+  if (result == ledger::Result::LEDGER_OK) {
+    if (data.empty()) {
+      result = ledger::Result::INVALID_NICEWARE_LIST;
```
bridiver: setting result here and returning doesn't do anything. Shouldn't we just pass `INVALID_NICEWARE_LIST` to the callback when data is empty?

bridiver: and shouldn't this also trigger OnRecoverWallet with the corresponding failure?

jasonrsadler: It seems that `INVALID_NICEWARE_LIST` was confusing the issue. Using `LEDGER_ERROR` instead and put in the OnRecoverWallet piece as well.

src/bat_client.cc
```diff
@@ -928,11 +928,45 @@ void BatClient::saveState() {
...
+}
+
+void BatClient::OnNicewareListLoaded(const std::string& pass_phrase,
+                                        ledger::Result result,
+                                        const std::string& data) {
+  if (result == ledger::Result::LEDGER_OK) {
+    std::vector<uint8_t> seed;
+    seed.resize(32);
+    size_t written = 0;
+    uint8_t result = braveledger_bat_helper::niceware_mnemonic_to_bytes(
```
bridiver: it's not a good idea to redefine result like this. We should keep it as a ledger::Result

jasonrsadler: Fixed

src/bat_client.cc
```diff
@@ -928,11 +928,45 @@ void BatClient::saveState() {
...
+      pass_phrase, seed, &written, braveledger_bat_helper::split(
+      data, DICTIONARY_DELIMITER));
+    continueRecover(result, &written, seed);
+  } else {
+    std::vector<braveledger_bat_helper::GRANT> empty;
+    ledger_->OnRecoverWallet(ledger::Result::LEDGER_ERROR, 0, empty);
```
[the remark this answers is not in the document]
jasonrsadler: Good catch.

src/test/niceware_partial_unittest.cc
jasonrsadler: Fixed and added.

src/test/niceware_partial_unittest.cc
```diff
@@ -0,0 +1,141 @@
...
+#include "testing/gtest/include/gtest/gtest.h"
+#include "components/grit/brave_components_resources.h"
+#include "ui/base/resource/resource_bundle.h"
+
+std::string getHexBytes(std::vector<uint8_t> seed) {
+  // std::vector<uint8_t> secretKey = braveledger_bat_helper::getHKDF(seed);
+  // std::vector<uint8_t> publicKey;
+  // std::vector<uint8_t> newSecretKey;
+  // braveledger_bat_helper::getPublicKeyFromSeed(
+  //   secretKey, publicKey, newSecretKey);
```
diracdeltas: can the commented-out lines be deleted?

src/bat_helper.cc
```diff
@@ -1936,4 +1936,32 @@ static bool ignore_ = false;
...
+  uint8_t niceware_mnemonic_to_bytes(const std::string& w,
+    std::vector<uint8_t>& bytes_out, size_t *written,
+    std::vector<std::string> wordDictionary) {
+    std::vector<std::string> wordList = split(toLowerCase(w),
+      WALLET_PASSPHRASE_DELIM);
+    std::vector<uint8_t> buffer(wordList.size() * 2);
+
+    for (uint8_t ix = 0; ix < wordList.size(); ix++) {
+      std::vector<std::string>::iterator it =
+        std::find(wordDictionary.begin(),
```
diracdeltas: since the wordlist is sorted, it's more efficient to use a binary search (`std::find` simply goes through each element of the wordlist and compares it to the value to-be-found). looks like https://en.cppreference.com/w/cpp/algorithm/binary_search would do it

src/test/niceware_partial_unittest.cc
jasonrsadler: Ah yes, thanks for catching

src/bat_helper.cc
jasonrsadler: I agree with the efficiency and I had originally looked at std::binary_search however it only tells us if the element exists (bool). We need to return the index and this doesn't have any functionality that provides that (I also didn't see anything similar to the js version of binary_search either, which prompted me to look at this beforehand).

diracdeltas: i see. sgtm as long as the performance difference isn't a big deal.

jasonrsadler: The human discernable time difference between this and bip39 was non-existent. Also, this is for backwards compatibility. Users would preferentially use the bip39 words.

diracdeltas:
lgtm as long as someone has gone through the tests and made sure that they pass

NejcZdovc:
@diracdeltas will test it

NejcZdovc:
tests are passing on core

Issue #124, opened 2018-10-02
Publishers list should not be fetched if rewards is not enabled


ryanml:

ryanml:
This was fixed with: https://github.com/brave-intl/bat-native-ledger/pull/127

PR #125, opened 2018-10-02
Only fetch list if rewards enabled

ryanml:
Fixes: #124

The previous fix concerning this: (https://github.com/brave-intl/bat-native-ledger/pull/123) made sure we didn't fetch publishers list more often than needed, but it introduced a case where the list would be fetched even if rewards was not enabled.

This tweak restores the original fetch, but adds an item `publisher_state_loaded_` so we can only refresh the list if the publisher state is loaded, so we don't get the overwrite problem.


NejcZdovc:
have an alternative implementation, will show you in the office as soon as build is done 😃

NejcZdovc:
closing in favour of https://github.com/brave-intl/bat-native-ledger/pull/127

PR #127, opened 2018-10-02
Only fetch publisher list if rewards enabled

NejcZdovc:


src/ledger_impl.cc
```diff
@@ -175,11 +175,13 @@ void LedgerImpl::OnLedgerStateLoaded(ledger::Result result,
                                         const std::string& data) {
   if (result == ledger::Result::LEDGER_OK) {
     if (!bat_client_->loadState(data)) {
-      result = ledger::Result::INVALID_LEDGER_STATE;
+      OnWalletInitialized(ledger::Result::INVALID_LEDGER_STATE);
```
bridiver: this looks like it changes the logic. Why would we call `LoadPublisherState` if loadState failed?

PR #128, opened 2018-10-02
Adds support for favicon icons

NejcZdovc:
- add support for twitch icons
- add fetch for twitch and youtube icons

### DIFF
--- a/src/bat_get_media.cc
+++ b/src/bat_get_media.cc
@@ -98,7 +98,7 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
 
   if (!publisher_info.get()) {
     std::string mediaURL = getMediaURL(mediaId, providerName);
-    if (YOUTUBE_MEDIA_TYPE == providerName) {
+    if (providerName == YOUTUBE_MEDIA_TYPE) {
       auto request = ledger_->LoadURL((std::string)YOUTUBE_PROVIDER_URL + "?format=json&url=" + ledger_->URIEncode(mediaURL),
         std::vector<std::string>(), "", "", ledger::URL_METHOD::GET, &handler_);
       handler_.AddRequestHandler(std::move(request),
@@ -112,18 +112,11 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
           _1,
           _2,
           _3));
-    } else if (TWITCH_MEDIA_TYPE == providerName) {
-      const std::string twitchMediaID =
-        mediaId.find(MEDIA_DELIMITER) != std::string::npos ?
-        mediaId :
-        braveledger_bat_helper::split(mediaId, MEDIA_DELIMITER)[0];
-      const std::string mediaUrl = getMediaURL(twitchMediaID, providerName);
-      std::unique_ptr<ledger::PublisherInfo> new_publisher_info(new ledger::PublisherInfo());
-      new_publisher_info->favicon_url = "";
-      new_publisher_info->url = mediaUrl + "/videos";
-      std::string id = providerName + "#author:" + twitchMediaID;
-      new_publisher_info->name = twitchMediaID;
-      new_publisher_info->id = id;
+    } else if (providerName == TWITCH_MEDIA_TYPE) {
+      int n = mediaId.length();
+      if (n == 0) {
+        return;
+      }
 
       ledger::TwitchEventInfo oldEvent;
       std::map<std::string, ledger::TwitchEventInfo>::const_iterator iter = twitchEvents.find(media_key);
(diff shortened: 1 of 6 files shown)


src/bat_get_media.cc
```diff
@@ -112,18 +112,11 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
...
-      const std::string mediaUrl = getMediaURL(twitchMediaID, providerName);
-      std::unique_ptr<ledger::PublisherInfo> new_publisher_info(new ledger::PublisherInfo());
-      new_publisher_info->favicon_url = "";
-      new_publisher_info->url = mediaUrl + "/videos";
-      std::string id = providerName + "#author:" + twitchMediaID;
-      new_publisher_info->name = twitchMediaID;
-      new_publisher_info->id = id;
+    } else if (providerName == TWITCH_MEDIA_TYPE) {
+      int n = mediaId.length();
+      if (n == 0) {
```
bbondy: nit: `if (mediaId.empty())`

src/bat_get_media.cc
```diff
@@ -141,26 +134,59 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
...
       updated_visit_data.provider = TWITCH_PROVIDER_NAME;
-      updated_visit_data.name = new_publisher_info->name;
-      updated_visit_data.url = new_publisher_info->url;
+
+      char char_array[n+1];
+      strcpy(char_array, mediaId.c_str());
+      if (strstr(char_array, "_vod_") != NULL) {
+        // VOD
+        std::vector<std::string> media_props = braveledger_bat_helper::split(mediaId, MEDIA_DELIMITER);
+        twitchMediaID = media_props[0];
```
bbondy: Only go into this code if media_props has a `!empty()`

src/bat_helper.cc
```diff
@@ -1883,9 +1883,9 @@ static bool ignore_ = false;
             iter = data.find("vod");
             if (iter != data.end()) {
               std::string idAddition(iter->second);
-              auto new_end = std::remove(idAddition.begin(), idAddition.end(), 'v');
-              id += "_vod_" + idAddition;
-              [&new_end] {}(); /*ignore [[nodiscard]]*/
+              if (idAddition.find('v') != std::string::npos) {
+                id += "_vod_" + braveledger_bat_helper::split(idAddition, 'v')[1];
```
bbondy: Not sure if it will always be the case that the split returns something with at least 2 elements here, maybe guard against invalid input.

src/bat_get_media.cc
```diff
@@ -293,15 +324,37 @@ void BatGetMedia::getPublisherFromMediaPropsCallback(const uint64_t& duration, c
                   _1,
                   _2,
                   _3));
+    return;
+  }
+
+  if (providerName == TWITCH_MEDIA_TYPE) {
+    std::string fav_icon = "";
```
bbondy: nit: no need for `= ""`

src/bat_helper.cc
NejcZdovc: isn't this guard enough `if (idAddition.find('v') != std::string::npos) {`?

src/bat_get_media.cc
```diff
@@ -141,26 +134,59 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
...
+        auto request = ledger_->LoadURL((std::string)TWITCH_PROVIDER_URL + "?json&url=" + ledger_->URIEncode(oembed_url),
+                                        std::vector<std::string>(), "", "", ledger::URL_METHOD::GET, &handler_);
+        handler_.AddRequestHandler(std::move(request),
+                                   std::bind(&BatGetMedia::getPublisherFromMediaPropsCallback,
+                                             this,
+                                             realDuration,
+                                             media_key,
+                                             providerName,
+                                             mediaUrl,
+                                             updated_visit_data,
```
bbondy: I think we need to make the parameters that are bound here not a reference in the signature of `getPublisherFromMediaPropsCallback`.

NejcZdovc: should we change this in this PR or have separate PR, because we use this logic elsewhere as well

src/bat_get_media.cc
```diff
@@ -141,24 +134,55 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
...
       ledger::VisitData updated_visit_data(visit_data);
-      updated_visit_data.favicon_url = new_publisher_info->favicon_url;
+      updated_visit_data.favicon_url = "";
       updated_visit_data.provider = TWITCH_PROVIDER_NAME;
-      updated_visit_data.name = new_publisher_info->name;
-      updated_visit_data.url = new_publisher_info->url;
+
+      char char_array[n+1];
+      strcpy(char_array, mediaId.c_str());
+      if (strstr(char_array, "_vod_") != NULL) {
```
SergeyZhukovsky: you can just use that `if (mediaId.find("_vod_") != std::string::npos)` instead of the above 3 lines and copying one string to another

src/bat_get_media.cc
```diff
@@ -141,24 +134,55 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId, const
...
+        // VOD
+        twitchMediaID = braveledger_bat_helper::split(mediaId, MEDIA_DELIMITER)[0];
```
SergeyZhukovsky: that place will crash in case of mediaId is empty or it doesn't contain MEDIA_DELIMITER as it will return an empty vector and you will try to access it's first element [0]. It's better to get that vector first and check on length.

src/bat_get_media.cc
NejcZdovc: done

src/bat_get_media.cc
NejcZdovc: done

src/bat_get_media.cc
```diff
@@ -293,15 +325,37 @@ void BatGetMedia::getPublisherFromMediaPropsCallback(const uint64_t& duration, c
...
+    std::string fav_icon;
+    braveledger_bat_helper::getJSONValue("author_thumbnail_url", response, fav_icon);
+    std::string author_name;
+    braveledger_bat_helper::getJSONValue("author_name", response, author_name);
+
+    std::string twitchMediaID = visit_data.name;
+    std::string id = providerName + "#author:" + twitchMediaID;
+
+    ledger::VisitData updated_visit_data(visit_data);
+    updated_visit_data.favicon_url = "https://" + providerName + "-" + twitchMediaID + "-brave.com";
```
diracdeltas: do these URLs ever get fetched?

src/bat_get_media.cc
```diff
@@ -317,17 +371,18 @@ void BatGetMedia::getPublisherInfoCallback(const uint64_t& duration, const std::
...
       publisher_id += publisherURL.substr(pos + 1);
     }
-    std::unique_ptr<ledger::PublisherInfo> publisher_info(new ledger::PublisherInfo());
-    publisher_info->name = publisherName;
-    publisher_info->url = mediaURL;
-    publisher_info->favicon_url = favIconURL;
-    publisher_info->id = publisher_id;
+
+    if (favIconURL.length() > 0) {
+      std::string favicon_key = "https://" + providerName + "-" + publisherName + "-brave.com";
```
diracdeltas: same as for https://github.com/brave-intl/bat-native-ledger/pull/128/files#diff-36c660d1cc9916df30542a8bbb1728faR341

src/bat_get_media.cc
diracdeltas: 2 potential issues:
1. attacker could register the domain name this is calling 
2. this leaks provider name and twitch ID to a network attacker via DNS

NejcZdovc: not this is not fetched is just ID that we use

diracdeltas: could you make this an invalid URL so that the above issues are not relevant if this ever gets fetched somewhere by accident?

diracdeltas: such that a check like https://github.com/brave/brave-core/pull/555/files#diff-a5d16be9c9b52a6df7b866b1d75d3f5fR1211 would return false

NejcZdovc: I tried first with invalid url but was not working

PR #131, opened 2018-10-08
Adds support for site banner

NejcZdovc:
implemented in https://github.com/brave/brave-core/pull/556

### DIFF
--- a/src/bat_helper.cc
+++ b/src/bat_helper.cc
@@ -1250,6 +1250,19 @@ static bool ignore_ = false;
 
   BATCH_PROOF::~BATCH_PROOF() {}
 
+  /////////////////////////////////////////////////////////////////////////////
+  SERVER_LIST_BANNER::SERVER_LIST_BANNER() {}
+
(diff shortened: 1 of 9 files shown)


src/bat_helper.cc
```diff
@@ -1552,6 +1565,49 @@ static bool ignore_ = false;
         SERVER_LIST item;
         item.verified = i[1].GetBool();
         item.excluded = i[2].GetBool();
+
+        SERVER_LIST_BANNER banner;
+
+        if (i[3].IsObject()) {
```
jasonrsadler: check if i.size goes to 3

src/bat_helper.cc
```diff
@@ -1552,6 +1565,49 @@ static bool ignore_ = false;
...
+          if (i[3].HasMember("title") && i[3]["title"].IsString()) {
+            banner.title_ = i[3]["title"].GetString();
+          } else {
+            banner.title_ = "";
```
jasonrsadler: No else clause is needed for these. They're empty strings by ctor.

PR #133, opened 2018-10-09
Auto-excluded publishers now show properly in Rewards panel.

### LINKED ISSUE #132: Remove exclude check
Remove exclude check in https://github.com/brave-intl/bat-native-ledger/blob/fa06a4b1153b5a1a87f1af8becfa0803e82bc6f3/src/bat_publishers.cc#L227 so that publisher is added even excluded. This way panel have publisher inside all the time

jasonrsadler:
Fixes #132 

Publishers that are set to auto-exclude now have details shown in rewards panel.

(Also corrected misspelling in unrelated function)

### DIFF
--- a/src/bat_publishers.cc
+++ b/src/bat_publishers.cc
@@ -230,18 +230,17 @@ void BatPublishers::saveVisitInternal(
                                                    visit_data.local_month,
                                                    visit_data.local_year));
 
-  if (isExcluded(publisher_info->id, publisher_info->excluded)) {
-    return;
-  }
-
-
   publisher_info->favicon_url = visit_data.favicon_url;
   publisher_info->name = visit_data.name;
   publisher_info->provider = visit_data.provider;
   publisher_info->url = visit_data.url;
-  publisher_info->duration += duration;
   publisher_info->visits += 1;
   publisher_info->category = ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE;
+  if (!isExcluded(publisher_info->id, publisher_info->excluded)) {
+    publisher_info->duration += duration;
+  } else {
+    publisher_info->duration = 0;
+  }
   publisher_info->score += concaveScore(duration);
   publisher_info->verified = isVerified(publisher_info->id);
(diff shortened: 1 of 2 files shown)


src/bat_publishers.cc
```diff
@@ -224,18 +224,15 @@ void BatPublishers::saveVisitInternal(
...
   publisher_info->favicon_url = visit_data.favicon_url;
   publisher_info->name = visit_data.name;
   publisher_info->provider = visit_data.provider;
   publisher_info->url = visit_data.url;
-  publisher_info->duration += duration;
   publisher_info->visits += 1;
-  publisher_info->category = ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE;
+  if (!isExcluded(publisher_info->id, publisher_info->excluded)) {
+    publisher_info->category = ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE;
+    publisher_info->duration += duration;
```
NejcZdovc: Maybe we should set duration to 0 if `isExcluded`

jasonrsadler: Done.

src/bat_publishers.cc
```diff
@@ -224,18 +224,17 @@ void BatPublishers::saveVisitInternal(
...
-
   publisher_info->favicon_url = visit_data.favicon_url;
   publisher_info->name = visit_data.name;
   publisher_info->provider = visit_data.provider;
   publisher_info->url = visit_data.url;
-  publisher_info->duration += duration;
   publisher_info->visits += 1;
-  publisher_info->category = ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE;
+  if (!isExcluded(publisher_info->id, publisher_info->excluded)) {
+    publisher_info->category = ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE;
```
NejcZdovc: Please move this out of if

jasonrsadler: Done.

PR #134, opened 2018-10-09
Add support for direct and recurring donations

kylehickinson:
Porting some JS changes Marshall had done in `bat-client`

core implementation: https://github.com/brave/brave-core/pull/611

### DIFF
--- a/src/bat_client.cc
+++ b/src/bat_client.cc
@@ -22,7 +22,7 @@ namespace braveledger_bat_client {
 BatClient::BatClient(bat_ledger::LedgerImpl* ledger) :
       ledger_(ledger),
       state_(new braveledger_bat_helper::CLIENT_STATE_ST()),
-      currentReconcile_(new braveledger_bat_helper::CURRENT_RECONCILE) {
+      currentReconciles_(new std::map<std::string, braveledger_bat_helper::CURRENT_RECONCILE>) {
   // Enable emscripten calls
   //braveledger_bat_helper::readEmscripten();
   initAnonize();
@@ -281,37 +281,36 @@ bool BatClient::isReadyForReconcile() {
   return true;
 }
 
-void BatClient::reconcile(const std::string& viewingId, const braveledger_bat_helper::RECONCILE_OPTIONS& options) {
-  currentReconcile_->viewingId_ = viewingId;
+void BatClient::reconcile(const std::string& viewingId, bool reccuring, const std::vector<braveledger_bat_helper::RECONCILE_DIRECTION>& directions) {
+  if (currentReconciles_->count(viewingId) > 0) {
+    LOG(ERROR) << "unable to reconcile with the same viewing id";
+    return;
+  }
+  
+  auto reconcile = braveledger_bat_helper::CURRENT_RECONCILE();
 
   double fee = .0;
-  for (auto direction : options.directions_) {
-    const auto amount = direction.amount_;
-    
-    if (direction.publisher_key_.empty()) {
+  for (const auto& direction : directions) {
+    if (direction.publisher_.name.empty()) {
       LOG(ERROR) << "reconcile direction missing publisher";
       return;
     }
     
     if (direction.currency_ != CURRENCY) {
-      LOG(ERROR) << "reconcile direction currency invalid for " << direction.publisher_key_;
+      LOG(ERROR) << "reconcile direction currency invalid for " << direction.publisher_.name;
       return;
     }
     
-    fee += amount;
+    fee += direction.amount_;
   }
   
-  if ((fee > 0) && (!options.immediate_)) {
-    const auto currency = state_->walletProperties_.altcurrency_;
-    if (currency != CURRENCY) {
-      LOG(ERROR) << "invalid recurring currency " << currency;
-      return;
-    }
-    fee += state_->walletProperties_.fee_amount_;
-  }
-  
-  currentReconcile_->fee_ = fee;
+  reconcile.viewingId_ = viewingId;
+  reconcile.fee_ = fee;
+  reconcile.directions_ = directions;
+  reconcile.reccuring_ = reccuring;
 
+  currentReconciles_->insert(std::make_pair(viewingId, reconcile));
+  
   auto request_id = ledger_->LoadURL(braveledger_bat_helper::buildURL((std::string)RECONCILE_CONTRIBUTION + state_->userId_, PREFIX_V2),
       std::vector<std::string>(), "", "",
       ledger::URL_METHOD::GET,
@@ -320,12 +319,14 @@ void BatClient::reconcile(const std::string& viewingId, const braveledger_bat_he
   handler_.AddRequestHandler(std::move(request_id),
                              std::bind(&BatClient::reconcileCallback,
                                        this,
+                                       viewingId,
                                        _1,
                                        _2,
                                        _3));
 }
 
-void BatClient::reconcileCallback(bool result,
+void BatClient::reconcileCallback(const std::string& viewingId,
+                                  bool result,
                                   const std::string& response,
                                   const std::map<std::string, std::string>& headers) {
   if (!result) {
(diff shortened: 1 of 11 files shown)


src/bat_helper.h
```diff
@@ -336,6 +336,24 @@ namespace braveledger_bat_helper {
...
+    int amount_;
+    std::string currency_;
+  };
+  
+  struct RECONCILE_OPTIONS {
+    RECONCILE_OPTIONS();
+    ~RECONCILE_OPTIONS();
+    
+    std::vector<RECONCILE_DIRECTION> directions_;
+    bool immediateP_;
```
NejcZdovc: please rename this to `immediate`

src/bat_helper.h
```diff
@@ -336,6 +336,24 @@ namespace braveledger_bat_helper {
...
     std::string amount_;
     std::string currency_;
+    double fee_;
+  };
+  
+  struct RECONCILE_DIRECTION {
+    RECONCILE_DIRECTION();
+    ~RECONCILE_DIRECTION();
+    
+    std::string publisher_;
```
NejcZdovc: please rename this to `publisher_key_`

NejcZdovc:
@kylehickinson where do we add publisher form options to contribution list of publishers?

NejcZdovc:
@kylehickinson can you please rebase

src/bat_client.cc
```diff
@@ -292,24 +319,26 @@ void BatClient::reconcile(const std::string& viewingId) {
...
                                   const std::string& response,
                                   const std::map<std::string, std::string>& headers) {
   if (!result) {
     // TODO errors handling
     return;
   }
   //currentReconcile_->viewingId_ = extraData.string1;
-  braveledger_bat_helper::getJSONValue(SURVEYOR_ID, response, currentReconcile_->surveyorInfo_.surveyorId_);
-  currentReconcile();
+  braveledger_bat_helper::getJSONValue(SURVEYOR_ID, response, currentReconciles_->at(viewingId).surveyorInfo_.surveyorId_);
```
kylehickinson: Whoops, missed using `GetReconcileById` here

NejcZdovc:
please don't merge it before I have brave-core PR that can test the whole flow

mrose17:
re: https://github.com/brave-intl/bat-native-ledger/pull/134#issuecomment-428996215

+1

src/bat_client.cc
NejcZdovc: fixed

src/bat_client.cc
```diff
@@ -178,9 +178,9 @@ void BatClient::registerPersonaCallback(bool result,
 }
 
 void BatClient::resetReconcileStamp() {
-  state_->reconcileStamp_ = braveledger_bat_helper::currentTime() + braveledger_ledger::_reconcile_default_interval;
-  // For testing (reconcile will happen 5min after wallet creation
-  //state_->reconcileStamp_ = braveledger_bat_helper::currentTime() + 5 * 60;
+  // state_->reconcileStamp_ = braveledger_bat_helper::currentTime() + braveledger_ledger::_reconcile_default_interval;
+  // For testing (reconcile will happen 3min after wallet creation
+  state_->reconcileStamp_ = braveledger_bat_helper::currentTime() + 3 * 60;
```
NejcZdovc: Don't forget to remove this one

src/bat_client.cc
```diff
@@ -329,18 +346,32 @@ void BatClient::reconcileCallback(const std::string& viewingId,
...
+
+  if (!result || reconcile.viewingId_.empty()) {
     // TODO errors handling
     return;
   }
-  //currentReconcile_->viewingId_ = extraData.string1;
-  braveledger_bat_helper::getJSONValue(SURVEYOR_ID, response, currentReconciles_->at(viewingId).surveyorInfo_.surveyorId_);
+
+  braveledger_bat_helper::getJSONValue(SURVEYOR_ID, response, reconcile.surveyorInfo_.surveyorId_);
+  bool success = SetReconcile(reconcile);
```
kylehickinson: Wouldn't these calls to `SetReconcile` fail, assuming you're trying to update the reconcile in `currentReconciles_` after changing `reconcile`. I don't see them being removed by calling `removeReconcileById` anywhere except `OnCompleteReconcile`

NejcZdovc: sure I can rename it. Without this change data was not saved so `surveyorId_` was empty in the next function

kylehickinson: Yeah I can see that now 😅 Better to do this rather than have `GetReconcileById` return a reference, correct?

NejcZdovc: I feel more confident in this approach and we update only when need and we don't need ref every time

NejcZdovc: This way flow is known

kylehickinson: OK sounds good

NejcZdovc:
@kylehickinson this is done if you would like to review

src/bat_publishers.cc
```diff
@@ -371,6 +371,26 @@ bool BatPublishers::getPublisherAllowVideos() const {
   return state_->allow_videos_;
 }
 
+void BatPublishers::synopsisNormalizerInternalContributre(ledger::PublisherInfoList* newList, bool saveData,
```
kylehickinson: nit/typo: `Contributre` → `Contributer`

src/bat_publishers.cc
```diff
@@ -454,49 +474,89 @@ void BatPublishers::synopsisNormalizer(const ledger::PublisherInfo& info) {
...
-      ledger::PUBLISHER_EXCLUDE::DEFAULT,
-      true,
-      currentReconcileStamp);
-  // TODO SZ: We pull the whole list currently, I don't think it consumes lots of RAM, but could.
-  // We need to limit it and iterate.
-  ledger_->GetPublisherInfoList(0, 0, filter, std::bind(&BatPublishers::topNInternal, this,
-          ballots, viewing_id, _1, _2));
+void BatPublishers::topN(const unsigned int& ballots, const std::string& viewing_id) {
+  const auto reconcile = ledger_->GetReconcileById(viewing_id);
+  if (reconcile.category_ == ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE) {
```
kylehickinson: Style-nit: Why not use a `switch` on `reconcile.category_`?

src/ledger_impl.cc
```diff
@@ -718,4 +785,70 @@ ledger::PublisherBanner LedgerImpl::GetPublisherBanner(const std::string& publis
...
+    for (auto &publisher : reconcile.list_) {
+      // TODO NZ remove when we completely switch to probi
+      const std::string probi = std::to_string((int)publisher.weight_) + "000000000000000000";
+      ledger_client_->SaveContributionInfo(probi, month, year, date, publisher.id_, category);
+    }
+
+    return;
+  }
+
+  bat_client_->removeReconcileById(viewing_id);
```
kylehickinson: Assuming we're always reconciling currently with any of the 3 supported publisher categories, this will never be called and `currentReconciles_` would keep growing?

src/bat_publishers.cc

src/bat_publishers.cc

src/ledger_impl.cc
NejcZdovc: added to error flow, so that list shouldn't grow anymore

src/ledger_impl.h
```diff
@@ -158,13 +163,25 @@ class LedgerImpl : public ledger::Ledger,
                             int year,
                             ledger::ReportType type,
                             const std::string& probi) override;
+
+  braveledger_bat_helper::CURRENT_RECONCILE GetReconcileById(const std::string& viewingId);
+  void RemoveReconcileById(const std::string& viewingId);
```
kylehickinson: Since you moved the reconciles to the state, this method can be removed :)

src/bat_client.h
```diff
@@ -45,7 +46,15 @@ class BatClient {
   double getContributionAmount() const;
   bool getAutoContribute() const;
   bool isReadyForReconcile();
-  void reconcile(const std::string& viewingId);
+  braveledger_bat_helper::CURRENT_RECONCILE GetReconcileById(const std::string& viewingId);
+  bool SetReconcile(const braveledger_bat_helper::CURRENT_RECONCILE& reconcile);
+  void removeReconcileById(const std::string& viewingId);
```
kylehickinson: Same as https://github.com/brave-intl/bat-native-ledger/pull/134/files#diff-aefa2765148ba607c414800210ab4586R168

src/ledger_impl.h
NejcZdovc: we still use it on error, so this is still needed

kylehickinson: Ah, I see it added now :) Gotcha

kylehickinson:
"Approved" but since I'm the PR starter I can't approve it

diracdeltas:
can tests be added

NejcZdovc:
@diracdeltas added to the list

PR #160, opened 2018-10-27
Grant checks are now performed every 24 hours.

jasonrsadler:
Core implementation: https://github.com/brave/brave-core/pull/751

Grant checks are now performed every 24 hours.

### DIFF
--- a/src/ledger_impl.cc
+++ b/src/ledger_impl.cc
@@ -35,7 +35,8 @@ LedgerImpl::LedgerImpl(ledger::LedgerClient* client) :
     last_pub_load_timer_id_(0u),
     last_reconcile_timer_id_(0u),
     last_prepare_vote_batch_timer_id_(0u),
-    last_vote_batch_timer_id_(0u) {
+    last_vote_batch_timer_id_(0u),
+    last_grant_check_timer_id_(0u) {
 }
 
 LedgerImpl::~LedgerImpl() {
@@ -250,6 +251,7 @@ void LedgerImpl::OnWalletInitialized(ledger::Result result) {
     initialized_ = true;
     LoadPublisherList(this);
     Reconcile();
+    RefreshGrant(false);
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/ledger_impl.cc
```diff
@@ -742,6 +744,42 @@ void LedgerImpl::RefreshPublishersList(bool retryAfterError) {
...
+  }
+  ledger_client_->SetTimer(start_timer_in, last_grant_check_timer_id_);
+}
+
+uint64_t LedgerImpl::retryRequestSetup(uint64_t min_time, uint64_t max_time) {
+  std::random_device seeder;
+  const auto seed = seeder.entropy() ? seeder() : time(nullptr);
+  std::mt19937 eng(static_cast<std::mt19937::result_type>(seed));
+
+  std::uniform_int_distribution <> dist(min_time, max_time);
```
NejcZdovc: I think we need to do DCHECK here in case you are sending in max time that is smaller then min time

src/ledger_impl.cc
```diff
@@ -565,7 +567,8 @@ void LedgerImpl::OnGrant(ledger::Result result, const braveledger_bat_helper::GR
   ledger::Grant grant;
 
   grant.promotionId = properties.promotionId;
-
+  RefreshGrant(result != ledger::Result::LEDGER_OK);
+  last_grant_check_timer_id_ = 0;
```
NejcZdovc: Is this intentional that you only set timer here to 0 and not before refreshGrant?

because in refreshGrant you have this line at the beginning

```
 if (last_grant_check_timer_id_ != 0) {
    return;
  }
```  

and timer id should be always set, based on the fact that should check every 24h

NejcZdovc: and when you get grant you should set new 24h cycle

jasonrsadler: No. This should have been one line up. Might have gotten caught in 'undo' before check in. Fixed.

src/ledger_impl.cc
jasonrsadler: Done.

src/ledger_impl.cc
jasonrsadler: Also, this resets the timer as GetGrant flows to OnGrant (OnGrant is the first non-const function to do this in)

PR #166, opened 2018-10-31
Rewards summary should now be cleared when wallet is restored

jasonrsadler:
Core implementation: https://github.com/brave/brave-core/pull/777

### DIFF
--- a/src/bat_publishers.cc
+++ b/src/bat_publishers.cc
@@ -722,6 +722,14 @@ bool BatPublishers::isEligibleForContribution(const ledger::PublisherInfo& info)
 
 }
 
+void BatPublishers::clearAllBalanceReports() {
+  if (state_->monthly_balances_.empty()) {
+    return;
+  }
+  state_->monthly_balances_.clear();
+  saveState();
+}
+
 void BatPublishers::setBalanceReport(ledger::PUBLISHER_MONTH month,
                                 int year,
                                 const ledger::BalanceReportInfo& report_info) {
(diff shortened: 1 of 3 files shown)


src/bat_publishers.cc
```diff
@@ -745,6 +753,7 @@ void BatPublishers::setBalanceReport(ledger::PUBLISHER_MONTH month,
 
   report_balance.total_ = total;
   state_->monthly_balances_[GetBalanceReportName(month, year)] = report_balance;
+
```
NejcZdovc: remove this line

src/ledger_impl.cc
```diff
@@ -593,7 +593,7 @@ void LedgerImpl::RecoverWallet(const std::string& passPhrase) const {
 
 void LedgerImpl::OnRecoverWallet(ledger::Result result, double balance, const std::vector<braveledger_bat_helper::GRANT>& grants) {
   std::vector<ledger::Grant> ledgerGrants;
-
+  LOG(ERROR) << "RECOVERING===========";
```
NejcZdovc: remove this log

src/ledger_impl.cc
```diff
@@ -604,6 +604,8 @@ void LedgerImpl::OnRecoverWallet(ledger::Result result, double balance, const st
     ledgerGrants.push_back(tempGrant);
   }
 
+  bat_publishers_->clearAllBalanceReports();
```
NejcZdovc: clear reports only if recover is successful

src/bat_publishers.cc
jasonrsadler: Done

src/ledger_impl.cc
jasonrsadler: Done

src/ledger_impl.cc
jasonrsadler: Done

PR #168, opened 2018-11-05
Splits contribution flow into separate file

NejcZdovc:
core implementation: https://github.com/brave/brave-core/pull/806

This PR moves all logic regarding contribution into separate class so that we have better abstraction. I tried to keep as much as possible the old logic, to prevent regressions.


NejcZdovc:
master (0.59) 0866b29
0.58.x 22b0484

PR #170, opened 2018-11-05
Grant fetch timestamp is now reset when no grants are available

jasonrsadler:
Core Implementation: https://github.com/brave/brave-core/pull/807

Also does not attempt retries on error since 404s are considered errors.

### DIFF
--- a/src/ledger_impl.cc
+++ b/src/ledger_impl.cc
@@ -573,7 +573,8 @@ void LedgerImpl::OnGrant(ledger::Result result, const braveledger_bat_helper::GR
 
   grant.promotionId = properties.promotionId;
   last_grant_check_timer_id_ = 0;
-  RefreshGrant(result != ledger::Result::LEDGER_OK);
+  RefreshGrant(result != ledger::Result::LEDGER_OK &&
+    result != ledger::Result::GRANT_NOT_FOUND);
   ledger_client_->OnGrant(result, grant);
 }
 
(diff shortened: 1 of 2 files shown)


src/ledger_impl.cc
```diff
@@ -573,7 +573,7 @@ void LedgerImpl::OnGrant(ledger::Result result, const braveledger_bat_helper::GR
 
   grant.promotionId = properties.promotionId;
   last_grant_check_timer_id_ = 0;
-  RefreshGrant(result != ledger::Result::LEDGER_OK);
+  RefreshGrant(false);
```
NejcZdovc: Can you give me some more background why this was changed? Why we don't want to retry on error?

jasonrsadler: If no grants are available it returns 404 error. We don't want to use retry interval if no grants are available, we would just check on the next 24 hour interval.

NejcZdovc: but what happens if there is server error and not 404?

jasonrsadler: We try again in 24 hours.

NejcZdovc: let's do server errors and our errors via retries and 404 via 24 cycle

jasonrsadler: No problem 😁

jasonrsadler: Done

PR #171, opened 2018-11-05
Splits state

NejcZdovc:
core implementation: https://github.com/brave/brave-core/pull/814

In this PR I moved client state into separate class so it's not coupled to bat_client anymore. Now all other classes can use it the same way via ledgerImpl.

This was done, because I will decoupled contribution from from bat_client into separate class. This will be a base for retry logic.

### DIFF
--- a/src/bat_client.cc
+++ b/src/bat_client.cc
@@ -20,50 +20,11 @@ using namespace std::placeholders;
 namespace braveledger_bat_client {
 
 BatClient::BatClient(bat_ledger::LedgerImpl* ledger) :
-      ledger_(ledger),
-      state_(new braveledger_bat_helper::CLIENT_STATE_ST()) {
-  // Enable emscripten calls
-  //braveledger_bat_helper::readEmscripten();
+      ledger_(ledger) {
   initAnonize();
 }
 
 BatClient::~BatClient() {
-}
-
-bool BatClient::loadState(const std::string& data) {
-  braveledger_bat_helper::CLIENT_STATE_ST state;
-  if (!braveledger_bat_helper::loadFromJson(state, data.c_str())) {
-    ledger_->Log(__func__, ledger::LogLevel::LOG_ERROR, {"Failed to load client state: ", data});
-    return false;
-  }
-
-  state_.reset(new braveledger_bat_helper::CLIENT_STATE_ST(state));
-
-  bool stateChanged = false;
-
-  // clear old reconciles
-  if (state_->batch_.size() == 0) {
-    state_->current_reconciles_ = {};
-    stateChanged = true;
-  }
-
-  // fix timestamp ms to s conversion
-  if (std::to_string(state_->reconcileStamp_).length() > 10) {
-    state_->reconcileStamp_ = state_->reconcileStamp_ / 1000;
-    stateChanged = true;
-  }
-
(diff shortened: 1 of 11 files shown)


src/bat_client.cc
```diff
@@ -181,98 +156,36 @@ void BatClient::registerPersonaCallback(bool result,
...
-    state_->masterUserToken_ = masterUserToken;
+    ledger_->SetMasterUserToken(masterUserToken);
     free((void*)masterUserToken);
   } else if (!braveledger_bat_helper::ignore_for_testing()) {
     ledger_->OnWalletInitialized(ledger::Result::REGISTRATION_VERIFICATION_FAILED);
     return;
   }
 
-  if (!braveledger_bat_helper::getJSONWalletInfo(response, state_->walletInfo_, state_->fee_currency_, state_->fee_amount_, state_->days_)) {
+  braveledger_bat_helper::WALLET_INFO_ST wallet_info = ledger_->GetWalletInfo();
```
bridiver: why do we get the wallet info here and then set it again below?

src/bat_client.cc
```diff
@@ -181,98 +156,36 @@ void BatClient::registerPersonaCallback(bool result,
...
-  // Should we use _reconcile_default_interval or state_->days_?
-  // In seconds
-
-  resetReconcileStamp();
 
+  ledger_->SetWalletInfo(wallet_info);
+  ledger_->SetCurrency(currency);
+  ledger_->SetContributionAmount(fee_amount);
+  ledger_->SetDays(days);
+  ledger_->SetBootStamp(braveledger_bat_helper::currentTime());
```
bridiver: maybe some of these last 3 methods can be combined? Like OnWalletInitialized could internally call SetBootStamp/ResetReconcileStamp? It just feels like bat_client has to know a lot about what is going on in ledger here

src/bat_client.cc
NejcZdovc: because we need existing data then it goes to getJSONWalletInfo adds additional data and then merged data is saved

src/ledger_impl.cc
```diff
@@ -939,4 +944,150 @@ void LedgerImpl::LogResponse(const std::string& func_name,
   ledger_client_->Log(ledger::LogLevel::LOG_RESPONSE, "[ END RESPONSE ]");
 }
 
+void LedgerImpl::ResetReconcileStamp() {
```
bridiver: I guess I was thinking that you would just add LedgerImpl::GetState() instead of proxying every method. This seems like it adds a lot of bloat to ledger by proxying all of them

bridiver: sure, but if it's just a straight proxy of the method call there is really no difference to the other classes between ledger->SetWhatever and ledger->GetState()->SetWhatever.

bridiver: if the ledger method encapsulated more logic with state then I think it would make more sense to do it this way

bridiver: for instance right now there will be a lot of unnecessary calls to save the state

NejcZdovc:
master (0.59) f4ed0e1
0.58.x f4ed0e14a2e238f137530a0ff6b804e25fa21dca
0.57.x 07b1739

PR #177, opened 2018-11-11
Twitch panel now shows favicon and panel displays correctly for tipping

jasonrsadler:
Core implementation: https://github.com/brave/brave-core/pull/888

For brave/brave-browser#1814

### DIFF
--- a/src/bat_get_media.cc
+++ b/src/bat_get_media.cc
@@ -59,10 +59,10 @@ void BatGetMedia::processMedia(const std::map<std::string, std::string>& parts, 
     return;
   }
   std::string mediaId = braveledger_bat_helper::getMediaId(parts, type);
-  BLOG(ledger_, ledger::LogLevel::LOG_DEBUG) << "Media Id: " << mediaId;
   if (mediaId.empty()) {
     return;
   }
+  BLOG(ledger_, ledger::LogLevel::LOG_DEBUG) << "Media Id: " << mediaId;
   std::string media_key = braveledger_bat_helper::getMediaKey(mediaId, type);
   BLOG(ledger_, ledger::LogLevel::LOG_DEBUG) << "Media key: " << media_key;
   uint64_t duration = 0;
@@ -144,10 +144,6 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId,
       uint64_t realDuration = getTwitchDuration(oldEvent, newEvent);
       twitchEvents[media_key] = newEvent;
 
-      if (realDuration == 0) {
-        return;
-      }
-
       std::string twitchMediaID = mediaId;
       std::string mediaUrl = getMediaURL(twitchMediaID, providerName);
 
@@ -163,7 +159,7 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId,
         }
 
         twitchMediaID = media_props[0];
-        mediaUrl = getMediaURL(twitchMediaID, providerName);
+        mediaUrl = TWITCH_VOD_URL + media_props[2];
         std::string oembed_url = (std::string)TWITCH_VOD_URL + media_props[media_props.size() - 1];
         updated_visit_data.name = twitchMediaID;
         updated_visit_data.url = mediaUrl + "/videos";
@@ -312,7 +308,8 @@ uint64_t BatGetMedia::getTwitchDuration(const ledger::TwitchEventInfo& oldEventI
 
 void BatGetMedia::onFetchFavIcon(const std::string& publisher_key,
                                  bool success,
-                                 const std::string& favicon_url) {
+                                 const std::string& favicon_url,
+                                 const std::string& name) {
   uint64_t currentReconcileStamp = ledger_->GetReconcileStamp();
   auto filter = ledger_->CreatePublisherFilter(publisher_key,
       ledger::PUBLISHER_CATEGORY::AUTO_CONTRIBUTE,
@@ -323,13 +320,15 @@ void BatGetMedia::onFetchFavIcon(const std::string& publisher_key,
       currentReconcileStamp);
   ledger_->GetPublisherInfo(filter,
       std::bind(&BatGetMedia::onFetchFavIconDBResponse,
-      this, _1, _2, favicon_url));
+      this, _1, _2, favicon_url, name));
 }
 
 void BatGetMedia::onFetchFavIconDBResponse(ledger::Result result,
                                            std::unique_ptr<ledger::PublisherInfo> info,
-                                           const std::string& favicon_url) {
+                                           const std::string& favicon_url,
+                                           const std::string& name) {
   if (result == ledger::Result::LEDGER_OK && !favicon_url.empty()) {
+    info->name = name;
     info->favicon_url = favicon_url;
 
     ledger_->SetPublisherInfo(std::move(info),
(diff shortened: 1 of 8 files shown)


src/bat_get_media.cc
```diff
@@ -163,7 +159,7 @@ void BatGetMedia::getPublisherInfoDataCallback(const std::string& mediaId,
         }
 
         twitchMediaID = media_props[0];
-        mediaUrl = getMediaURL(twitchMediaID, providerName);
+        mediaUrl = "https://www.twitch.tv/videos/" + media_props[2];
```
NejcZdovc: please use static value for the url

src/bat_get_media.cc
```diff
@@ -384,20 +383,28 @@ void BatGetMedia::getPublisherFromMediaPropsCallback(const uint64_t& duration,
...
     braveledger_bat_helper::getJSONValue("author_thumbnail_url", response, fav_icon);
+    std::string author_url;
+    braveledger_bat_helper::getJSONValue("author_url", response, author_url);
     std::string author_name;
     braveledger_bat_helper::getJSONValue("author_name", response, author_name);
-
+    std::string media_url(mediaURL);
     std::string twitchMediaID = visit_data.name;
+    if (twitchMediaID == "twitch") {
+      twitchMediaID = extractData(author_url, "https://www.twitch.tv/", "/");
```
NejcZdovc: please use static value

src/bat_helper.cc
```diff
@@ -2259,7 +2259,7 @@ static bool ignore_ = false;
             if (iter != data.end()) {
               std::string idAddition(iter->second);
               if (idAddition.find('v') != std::string::npos) {
-                id += "_vod_" + braveledger_bat_helper::split(idAddition, 'v')[1];
+                id = "_vod_" + braveledger_bat_helper::split(idAddition, 'v')[1];
```
NejcZdovc: why was + removed here, now provider is missing

src/bat_get_media.cc
```diff
@@ -819,6 +849,20 @@ std::string BatGetMedia::getYoutubeMediaKeyFromUrl(
...
 }
 
+std::string BatGetMedia::getTwitchMediaKeyFromUrl(
+  const std::string& provider_type,
+  const std::string& id,
+  const std::string& url) {
+  if (id == "twitch") {
+    return std::string();
+  }
+  if (url.find("twitch.tv/videos") != std::string::npos) {
```
diracdeltas, line 859: this would match non-twitch URLs like `https://mysite.com/twitch.tv/videos`

src/bat_get_media.cc
```diff
@@ -809,6 +828,17 @@ std::string BatGetMedia::getYoutubeMediaIdFromUrl(const ledger::VisitData& visit
   return std::string();
 }
 
+std::string BatGetMedia::getTwitchMediaIdFromUrl(
+  const ledger::VisitData& visit_data) {
+  std::string mediaId =
+    extractData(visit_data.url, "twitch.tv/", "/");
```
diracdeltas, line 834: is visit_data.url gauranteed to be a twitch domain? (ex: not `https://othersite.com/twitch.tv/path`)

src/bat_get_media.cc
```diff
@@ -839,6 +883,8 @@ std::string BatGetMedia::extractData(const std::string& data,
       match = data.substr(startPos, endPos - startPos);
     } else if (endPos != std::string::npos) {
       match = data.substr(startPos, endPos);
+    } else {
+      match = data.substr(startPos, std::string::npos);
```
diracdeltas, line 887: add a unit test case for this?

src/bat_get_media.cc
```diff
@@ -848,4 +894,76 @@ std::string BatGetMedia::getNameFromChannel(const std::string& data) {
...
+  const std::string& page_blob,
+  ledger::PUBLISHER_MONTH month,
+  int year, const std::string& url) {
+  std::string userId;
+  std::string faviconUrl = parseTwitchData(userId, page_blob);
+  std::string name(userId);
+  if (!userId.empty()) {
+    std::transform(userId.begin(), userId.end(), userId.begin(), ::tolower);
+    std::string mediaKey;
+    if (url.find("twitch.tv/videos") != std::string::npos) {
```
diracdeltas, line 907: same as https://github.com/brave-intl/bat-native-ledger/pull/177/files#r239629859

src/bat_get_media.cc
```diff
@@ -848,4 +894,76 @@ std::string BatGetMedia::getNameFromChannel(const std::string& data) {
...
+    "<div class=\"tw-border-radius-medium tw-overflow-hidden\">"
+    "<img class=\"tw-avatar__img tw-image\" alt=\"", "\"");
+  return extractData(page_blob,
+    "<figure class=\"tw-avatar tw-avatar--size-36\">"
+    "<div class=\"tw-border-radius-medium tw-overflow-hidden\">"
+    "<img class=\"tw-avatar__img tw-image\" alt=\"" + userId + "\" src=\"",
+    "\"");
+}
+
+bool BatGetMedia::isPublisher(const std::string& page_blob) {
```
diracdeltas: should this first verify that the associated URL is at the Twitch origin before calling `parseTwitchData`?

src/bat_get_media.cc
jasonrsadler: `visit_data.url` originates from chrome's `tab.url`. This line is unreachable if `visit_data.url` is not `twitch.tv`.

src/bat_get_media.cc
jasonrsadler: Same as above. In this case `url` was provided from `visit_data.url`

src/bat_get_media.cc
jasonrsadler: Are we doing tests for private functions?

src/bat_get_media.cc
jasonrsadler: `url` is provided from here: https://github.com/brave/brave-core/blob/9e55b9f2ae809f58efda3296d3cb195109d97bd7/components/brave_rewards/resources/extension/brave_rewards/background/reducers/rewards_panel_reducer.ts#L29 and is verified here: https://github.com/brave/brave-core/blob/9e55b9f2ae809f58efda3296d3cb195109d97bd7/components/brave_rewards/resources/extension/brave_rewards/background/reducers/rewards_panel_reducer.ts#L136

Does this suffice?

src/bat_get_media.cc
jasonrsadler: If you are referring to `page_blob`, twitch origin is verified from https://github.com/brave/brave-core/blob/9e55b9f2ae809f58efda3296d3cb195109d97bd7/components/brave_rewards/resources/extension/brave_rewards/background/reducers/rewards_panel_reducer.ts#L136 
and 
https://github.com/brave/brave-core/blob/9e55b9f2ae809f58efda3296d3cb195109d97bd7/components/brave_rewards/resources/extension/brave_rewards/background/reducers/rewards_panel_reducer.ts#L144

src/bat_get_media.cc
diracdeltas: it was not obvious to me that every call to these functions would be wrapped inside a check for the twitch origin, but if that's always going to be the case, sgtm

src/bat_get_media.cc
diracdeltas: i see; so this will only ever be called with a `page_blob` that the caller has verified to be a twitch page? if that is the case, can this be renamed to something Twitch-specific like `hasTwitchPublisherInfo` ? similarly https://github.com/brave-intl/bat-native-ledger/pull/177/files/f2c5eff98bf243ed9bb60c98cb51f9a9e975a142#diff-e8e610d2ae7b879e83dd444090f35aa0R1064 could be `evaluateTwitchPublisherBlob`

otherwise it's hard for the reviewer to know what the code is supposed to do unless they trace all the callers

jasonrsadler: Agreed. Will change.

src/bat_get_media.cc
diracdeltas: 👍 didn't notice this was private

src/bat_get_media.cc
jasonrsadler: Done.

diracdeltas:
lgtm now (though i haven't functionally tested it)

PR #184, opened 2018-11-21
Support publisher branded site banners

emerick:
Related to brave/brave-core#939

Currently we are not displaying publisher branded images in our site banners, due to WebUI disallowing external network requests. This updates the URLs to use our custom `chrome://rewardsres` data source to fulfill the resource requests instead.


emerick:
Renamed `chrome://rewardsres` to `chrome://rewards-image` for clarity.

PR #185, opened 2018-11-26
Added isolated testing support

### LINKED ISSUE #126: Add testing support for bat-native-ledger
Some tests currently exist in brave-core that tests functionality in bat-native-ledger. Tests should be set up to run in bat-native-ledger independently.

tmancey:
Fixes https://github.com/brave-intl/bat-native-ledger/issues/126

### DIFF
--- a/src/bat_publishers.cc
+++ b/src/bat_publishers.cc
@@ -864,10 +864,16 @@ void BatPublishers::getPublisherBanner(const std::string& publisher_id,
 
       banner.title = values.banner.title_;
       banner.description = values.banner.description_;
-      banner.background = values.banner.background_;
-      banner.logo = values.banner.logo_;
       banner.amounts = values.banner.amounts_;
       banner.social = values.banner.social_;
+
+      // WebUI must not make external network requests, so map
+      // external resopurces to chrome://rewards-image and handle them
+      // via our custom data source
+      if (!values.banner.background_.empty())
+        banner.background = "chrome://rewards-image/" + values.banner.background_;
+      if (!values.banner.logo_.empty())
+        banner.logo = "chrome://rewards-image/" + values.banner.logo_;
     }
   }
 
@@ -888,7 +894,6 @@ void BatPublishers::getPublisherBanner(const std::string& publisher_id,
                                       _2);
 
   ledger_->GetPublisherInfo(filter, callbackGetPublisher);
-
 }
 
 void BatPublishers::onPublisherBanner(ledger::PublisherBannerCallback callback,


src/bat_native_ledger.cc
```diff
@@ -0,0 +1,21 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+#include <assert.h>
+
+#include "bat_native_ledger.h"
+#include "mock_ledger_client.h"
+#include "bat/ledger/ledger.h"
+
+int main() {
+  auto *mock_ledger_client = new bat_ledger::MockLedgerClient();
```
jasonrsadler: Could we use `std::unique_ptr` here to do the GC handling?

README.md
```diff
@@ -1 +1,64 @@
...
+make test
+```
+
+You can also add the `ARGS="-V"` command line argument for verbose output
+
+```
+cd build
+
+make test ARGS="-V"
+```
```
jasonrsadler: Should mention that existing installs of `cmake` should be 3.13.0 or build will fail with at BAT Anonize with 
```
BAT Native Anonize dependency
CMake Error at CMakeLists.txt:173 (add_compile_definitions):
  Unknown CMake command "add_compile_definitions".
```
(originally tried with existing 3.11.1)

README.md
```diff
@@ -1 +1,64 @@
...
+cd build
+
+cmake ..
+```
+
+#### Build executable
+
+```
+cd build
```
jasonrsadler: Remove this line since already in the build directory (or specify that command is done in build directory)

README.md
```diff
@@ -1 +1,64 @@
...
+```
+
+#### Debugging
+
+You can attach a debugger to the `batnativeledger` executable which can be found in the `build` folder
+
+#### Isolated Google Tests
+
+```
+cd build
```
jasonrsadler: remove this line (same as above)

README.md
```diff
@@ -1 +1,64 @@
...
+
+make test
+```
+
+You can also add the `ARGS="-V"` command line argument for verbose output
+
+```
+cd build
```
jasonrsadler: remove this line (same as above)

README.md
bridiver: I think we want to dump cmake anyway as soon as we transition the tests over to gn

README.md
tmancey: 👍

README.md
tmancey: 👍

tmancey:
One of the tests do fail right now as it is just an example to show passing and failing tests ready for more tests to be added

jasonrsadler:
I'm showing only 1 test total. But I do see what you're saying if I run it with verbosity. I can confirm that tests do fail if they should. I would recommend setting up to have tests passing.

bridiver:
I'm confused about why we are using cmake for all this stuff instead of integrating into the chromium tests. These libraries do not have to buildable or testable outside of brave-core and we want to be able to run the tests along with brave-core.

bridiver:
we should use the mocks for the brave-core tests, but adding cmake build files, especially when they have to include the individual source files for things like anonize seems like a maintenance nightmare

kylehickinson:
Just popping in to get some discussion on this:

> @bridiver: These libraries do not have to buildable or testable outside of brave-core and we want to be able to run the tests along with brave-core.

This library and its dependencies in particular do have to be at least buildable outside of brave-core for the iOS app. Since it will have to be built without a brave-core/chromium backing, we won't be able to use gn the same way it currently is used in brave-core.

Not sure if there's been any prior discussion/plans around this

bridiver:
@kylehickinson we can do builds inside brave-core that generate a standalone lib that can be used in ios

bridiver:
just want to be clear that building the lib inside brave-core does not mean that the build would only be available for brave core. We already have a standalone lib output that gets generated and can be built independently of the browser

mrose17:
@bridiver & @tmancey - the reason that cmake is there is because the underlying anonize2 library requires the relic toolkit, and the relic toolkit requires cmake. i'm not happy about using cmake at all, but that train left the station 2 years ago. after we ship FCS, we can investigate replacing the anonize2 library dependency on cmake to a chromium framework. i'd be happy with seeing that happen, as it will reduce maintenance issues, among other things. but the time to do that is after FCS, not now.

bridiver:
@mrose17 we've been building anonize with gn for a while now. @gdregalo fixed the VS studio build

mrose17:
apparently i missed something. what exactly is the "ask" here? if it compiles on all platforms, great! thanks.

bridiver:
the ask is to not use cmake and put the testing support in gn

mrose17:
My hope is that this won't take more than a day, so let's see if we can get this done tomorrow.

@tmancey - you have the token on this.

thanks!

mrose17:
groovy!

PR #186, opened 2018-11-28
Adding provider to ledger::PublisherBanner

ryanml:
Related issue: https://github.com/brave/brave-browser/issues/2238

Test with core integration: https://github.com/brave/brave-core/pull/976

Adds provider attribute to `PublisherBanner`. This attribute is set from a `PublisherInfo` object in `BatPublishers::onPublisherBanner`

PR #193, opened 2018-12-03
Revert "Adding provider to ledger::PublisherBanner"

jasonrsadler:
Reverts brave-intl/bat-native-ledger#186

Should not go into 0.59.x until uplift requested


jasonrsadler:
Invalid

PR #194, opened 2018-12-03
Adds publisherKey to OnExcludedSitesChanged callback

ryanml:
This is so that the publisher key of the publisher whose exclude attribute was updated can be reported back to the UI

core implementation: https://github.com/brave/brave-core/pull/1010


NejcZdovc:
master (0.60) ab50d0e
0.59.x d5602b1

PR #197, opened 2018-12-05
Fixes publishers that where included

NejcZdovc:
We need to include all publishers except publishers that are excluded directly.

core implementation: https://github.com/brave/brave-core/pull/1023


tmancey (approved):

NejcZdovc:
master (0.60) 042f625
0.59.x d0a3c7b
0.58.x a208050

PR #199, opened 2018-12-06
Adding percent filter to PublisherInfoFilter

ryanml:
Core implementation: https://github.com/brave/brave-core/pull/1032

Adds a filter property for publisher percentage, currently we only have the need to do by 0

### DIFF
--- a/include/bat/ledger/publisher_info.h
+++ b/include/bat/ledger/publisher_info.h
@@ -52,6 +52,11 @@ LEDGER_EXPORT enum PUBLISHER_EXCLUDE_FILTER {
   FILTER_ALL_EXCEPT_EXCLUDED = 3
 };
 
+LEDGER_EXPORT enum PUBLISHER_PERCENT_FILTER {
+  FILTER_ALL_PERCENT = -1,
+  FILTER_ZERO = 0
+};
+
 LEDGER_EXPORT struct PublisherInfoFilter {
   PublisherInfoFilter();
   PublisherInfoFilter(const PublisherInfoFilter& filter);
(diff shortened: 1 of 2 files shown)


NejcZdovc:
~~@ryanml please add corresponding core PR~~

include/bat/ledger/publisher_info.h
```diff
@@ -52,6 +52,10 @@ LEDGER_EXPORT enum PUBLISHER_EXCLUDE_FILTER {
   FILTER_ALL_EXCEPT_EXCLUDED = 3
 };
 
+LEDGER_EXPORT enum PUBLISHER_PERCENT_FILTER {
+  FILTER_ZERO = 0
```
NejcZdovc, line 57: I think that we need any or all as well so that can be used as default in constructor

ryanml: @NejcZdovc is this necessary or just a nice to have for the future? I am only using this constructor: https://github.com/brave/brave-core/pull/1028

NejcZdovc: you probably linked to the wrong PR. which constructor?

ryanml: My mistake, this one here: https://github.com/brave-intl/bat-native-ledger/blob/03fe3bf89e500cc2f96d627f8fede77a432a96bf/src/bat/ledger/ledger.cc#L93

jasonrsadler: @NejcZdovc It appears that if the percent filter isn't used or is non-zero, we get any and all.

ryanml: Added anyways :)

